PR #2, opened 2026-09-02
ci: fix the three shellcheck findings actionlint reported, two of them real

JonoPrest:
The `workflows are sound` step added in #1 failed on its first run against
`main` ([run 33619992830](https://github.com/JonoPrest/nits/actions/runs/33619992830)).

It passed locally because **shellcheck was not installed here**, so actionlint
silently skipped its shellcheck integration — my local gate was weaker than the
CI one, which is the same shape of mistake as validating that a YAML file
parses and concluding a step changed. shellcheck is installed now, and
`actionlint .github/workflows/*.yml` reproduces CI locally.

Two of the three findings were defects, not style.

## `SC2015` at release.yml:339 — a formula with no checksums passed validation

```bash
grep -oE 'sha256 "[^"]*"' formula.rb | grep -vqE 'sha256 "[0-9a-f]{64}"' \
  && { echo '::error::formula has a malformed sha256'; exit 1; } || true
```

With `pipefail`, the pipeline fails when the *first* grep matches nothing, so
`&&` is skipped and `|| true` swallows the failure. The check therefore reported
success on a formula containing zero `sha256` lines — the exact input it exists
to reject. Demonstrated before the fix:

```
$ bash c1.sh            # formula.rb has no sha256 lines
CHECK-PASSED (formula has zero sha256 lines!)
exit=0
```

Now it collects the digests, requires four, and validates each:

```
::error::formula has 0 sha256 entries, expected 4        # zero entries
::error::formula has a malformed sha256: 'zzz'           # four, one bad
entries: 4 / HAPPY-PATH PASSES                           # real rendered formula
```

## `SC2015` at release.yml:397 — a real mismatch reported nothing

```bash
pk=$(grep -oE "..." out/PKGBUILD | grep -oE '[0-9a-f]{64}')
```

Under `set -e` the assignment aborts the step when the greps match nothing, so
the `::error::` line naming the file and architecture never ran — a genuine
PKGBUILD/.SRCINFO mismatch surfaced as a bare non-zero exit. The greps now
tolerate no match and the `if` reports which side was missing:

```
::error::PKGBUILD/.SRCINFO sha256 mismatch for x86_64 (PKGBUILD=none SRCINFO=none)
```

## `SC2129` at release.yml:71 — style only

The plan step's individual `>> "$GITHUB_OUTPUT"` redirects are grouped, and the
`[ ... ] && echo` notice becomes an `if`.

I checked whether that `&&` was a latent `set -e` abort — it is not. Bash exempts
the command before the final `&&` in an AND-list, and more commands follow it:

```
$ bash -c 'set -euo pipefail; [ absent = at_head ] && echo notice; echo REACHED-END'
REACHED-END
exit=0
```

So the plan step would not have failed on a fresh release. The `if` is for
clarity, not a fix.

## Testing

`actionlint .github/workflows/*.yml` clean with shellcheck present. Each
rewritten check verified against the input the old one mishandled, and the two
formula checks verified against a real rendered formula so the happy path still
passes.

Unrelated but worth recording: the first dry run of the release pipeline
([33621278920](https://github.com/JonoPrest/nits/actions/runs/33621278920)) has
`plan` green and 5 of 6 build targets green, so the corrected runner labels
resolve.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,16 +73,19 @@ jobs:
           cargo xtask release-plan --package '${{ inputs.package }}' | tee plan.json
           version=$(jq -r .version plan.json)
           tag=$(jq -r .tag plan.json)
-          echo "version=$version" >> "$GITHUB_OUTPUT"
-          echo "tag=$tag" >> "$GITHUB_OUTPUT"
-          echo "dir=$(jq -r .dir plan.json)" >> "$GITHUB_OUTPUT"
-          # Space-separated, so the build job can loop over it in shell.
-          echo "binaries=$(jq -r '.binaries | join(" ")' plan.json)" >> "$GITHUB_OUTPUT"
-          # "<name>:<dir>" pairs, one system package per binary shipped.
-          echo "packages=$(jq -r '[.packages[] | "\(.name):\(.dir)"] | join(" ")' plan.json)" >> "$GITHUB_OUTPUT"
-          echo "tag_state=$(jq -r .tag_state plan.json)" >> "$GITHUB_OUTPUT"
-          [ "$(jq -r .tag_state plan.json)" = at_head ] && \
+          {
+            echo "version=$version"
+            echo "tag=$tag"
+            echo "dir=$(jq -r .dir plan.json)"
+            # Space-separated, so the build job can loop over it in shell.
+            echo "binaries=$(jq -r '.binaries | join(" ")' plan.json)"
+            # "<name>:<dir>" pairs, one system package per binary shipped.
+            echo "packages=$(jq -r '[.packages[] | "\(.name):\(.dir)"] | join(" ")' plan.json)"
+            echo "tag_state=$(jq -r .tag_state plan.json)"
+          } >> "$GITHUB_OUTPUT"
+          if [ "$(jq -r .tag_state plan.json)" = at_head ]; then
             echo "::notice::tag $tag already exists on this commit — resuming a partial release"
+          fi
           if [ "$(jq -r .collision plan.json)" = "true" ]; then
             reason=$(jq -r .collision_reason plan.json)
             echo "::error::$reason — bump the version in Cargo.toml before releasing."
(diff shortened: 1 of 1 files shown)


JonoPrest (commented):
Review by gpt-5.6-sol via Codex. 0 findings: 0 P1, 0 P2, 0 P3.

The grouped output write preserves behavior. Both rewritten validators now fail with actionable diagnostics on the previously swallowed/no-match cases, while their happy paths still accept the renderer output shape. I also checked zero-count, malformed, missing, unequal, and valid synthetic cases; the CI shellcheck and actionlint steps are green. The broader Rust job is still running.

JonoPrest:
Scope note: added 5cdc50d to this PR rather than opening a third one, since it touches the same `release.yml` and stacking would conflict.

The first release dry run ([33621278920](https://github.com/JonoPrest/nits/actions/runs/33621278920)) failed, and it is a one-line bug in the same workflow:

```
rpm --import <(gpg --export --armor "$GPG_KEY_ID")
error: /dev/fd/63: import read failed(0).
```

`rpm --import` reads its argument as a seekable file; a process substitution gives it a pipe. Fixed by exporting to `$RUNNER_TEMP` first. The verification loop also only surfaced a non-zero exit, so it now names the file and prints the failing `rpm --checksig` output.

Confirmed working on the runner before that point, which was the part I could not test locally: all six build targets, cargo-deb's `target/release/` handling under `--target` with `--no-build`, per-binary packaging of both `nits` and `nitsd`, and `rpmsign` with a loopback pinentry against the release key.

JonoPrest (commented):
Follow-up review of 5cdc50d by gpt-5.6-sol via Codex. 0 new findings.

The public key is now exported to a seekable runner-temp file before import, and the signature loop retains pipeline failure semantics while reporting the failing RPM and its checksig output. The Ubuntu RPM configuration uses a per-user database, so import and verification operate on the same keyring without sudo. CI shellcheck and actionlint are green; the broader Rust job is still running.

PR #3, opened 2026-09-02
ci: split into independent jobs, fix cache handling, check ReScript formatting

JonoPrest:
CI took **8.5 minutes** against a 4.3-minute baseline because of a total cache
miss (`No cache found.`): clippy 14s→93s, nextest 85s→178s, bench 60s→144s.
After the split, the cold run measured **4m16s**.

> This description was rewritten after review. Several things it originally
> proposed were wrong and were removed rather than kept — see the threads.

## Why the cache was empty

Caches are branch-scoped — a branch reads its own or its base branch's. All the
warm caches belonged to `release-pipeline`. Once that merged, main's first CI
run failed at actionlint, so main saved nothing, and every branch cut from main
built cold.

**The fix is the split itself, not a cache setting.** `shellcheck`/`actionlint`
now live in a `scripts` job with no Rust toolchain, so a workflow-lint failure
can no longer stop the Rust jobs from publishing their caches. That is what
makes the original failure mode impossible.

I first tried `cache-on-failure: true`. That was wrong and is **not** in the
final diff: `rust-cache` skips saving on an exact-key hit
(`if (isCacheUpToDate()) return` in `save.ts`) and GitHub cache keys are
immutable, so a run failing midway would save a *partial* cache under a fresh
exact key that no later run could ever replace — strictly worse than a cold
build.

`save-if` restricts cache writes to main. Not because PR-scoped caches go
unused — later runs of the same PR do restore them, and this repository has
entries whose `last_accessed_at` postdates their creation — but because at
roughly 1.1 GiB each they are short-lived duplicates that can evict the
broadly reusable main cache under GitHub's shared per-repo quota.

## Independent jobs

The single `rust` job compiled the workspace three times in sequence — clippy's
check artifacts, nextest's dev-with-codegen, `cargo bench`'s release — and each
gate hid the next one's result.

| Job | Needs |
| --- | --- |
| `scripts` | no Rust toolchain |
| `fmt` | rustfmt only — no cache, no web bundle |
| `clippy` | compiles the tauri and web crates, so needs `web-bundle` |
| `wasm` | nothing embedded: `nits-protocol` and `nits-client-core` contain no `include_dir!` |
| `test` | `nextest --workspace`, so needs `web-bundle` too, plus the fixture freshness check |
| `ui` | pnpm only |

No explicit `key:` inputs — `rust-cache` appends `GITHUB_JOB` when
`add-job-id-key` is set, which is the default (`config.ts`). The release
workflow's existing `v0-rust-plan-…` and `v0-rust-crates-io-…` keys are that
behaviour with no `key` set.

## Benchmarks

They are a regression trigger, not a correctness gate, and cost a full
release-profile build. They move to `.github/workflows/bench.yml` and run on
**main pushes or manual dispatch only** — never in PR CI.

An earlier revision offered a `perf` label opt-in. Review found that `labeled`
is not a default `pull_request` activity type, so it never worked; and once
fixed, that any *later* label would restart an opted-in run, while
workflow-level concurrency would cancel an in-flight benchmark before the job's
`if` was evaluated. The repo owner then decided benchmarks should not run on
PRs at all, which removes that surface entirely. Job-level concurrency is kept
so back-to-back main pushes cancel the older run.

## Actions off Node 20

Every run was annotating `Node.js 20 is deprecated`. Six actions were on a major
whose `action.yml` declares `using: node20`:

```
actions/checkout            v4 -> v7   (13x)
actions/download-artifact   v4 -> v8   (5x)
actions/upload-artifact     v4 -> v7   (2x)
actions/setup-node          v4 -> v7   (2x)
pnpm/action-setup           v4 -> v6   (2x)
softprops/action-gh-release v2 -> v3   (1x)
```

`Swatinem/rust-cache@v2` already declares node24; `taiki-e/install-action` is
composite. These are major bumps, so I read each new `action.yml` rather than
assuming inputs survived — `fetch-depth`, `merge-multiple`,
`cache-dependency-path`, `generate_release_notes` and the rest all still exist,
and `ui/package.json` has no `packageManager` field for pnpm v6 to conflict
with.

## ReScript formatting

`pnpm rescript format --check` gates the `ui` job. 28 files needed formatting;
that is its own commit (`49ed7c9`), pure `rescript format` output, so the
enforcing commit carries no reformatting. 45 modules compile and all 477 tests
pass unchanged. Negative-tested rather than trusted: with a deliberately
misformatted `App.res` it exits 1, and 0 once restored.

## A note on the linter catching me

I first wrote the shared cache options as a YAML anchor with a `<<:` merge key.
`yaml.safe_load` parsed it happily; GitHub Actions does not support merge keys,
and the actionlint step added in #2 caught it.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -11,17 +11,27 @@ on:
     # explicitly, adding the label would not trigger anything.
     types: [opened, reopened, synchronize, labeled]
 
-concurrency:
-  group: bench-${{ github.ref }}
-  cancel-in-progress: true
-
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
   bench:
     name: benchmark triggers
-    if: github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'perf')
+    # On `labeled`, gate on the label that was just added: otherwise, once a PR
+    # carries `perf`, adding any *other* label satisfies a whole-label-set check
+    # and starts another release build. Every other activity type checks the
+    # full set, since no single label is being added.
+    if: >-
+      github.event_name == 'push'
+      || (github.event.action == 'labeled' && github.event.label.name == 'perf')
+      || (github.event.action != 'labeled'
+          && contains(github.event.pull_request.labels.*.name, 'perf'))
+    # Job-level, not workflow-level: workflow concurrency is applied before a
+    # job's `if` is evaluated, so a run that is going to skip would still cancel
+    # a benchmark already in flight.
+    concurrency:
+      group: bench-${{ github.ref }}
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,33 @@
+# Benchmarks are a regression *trigger*, not a correctness gate, and cost a
+# whole release-profile build of the workspace. They live in their own workflow
+# so that opting a PR in does not re-run every other CI job, and so `labeled`
+# can be an activity type here without making every label change re-run CI.
+name: bench
+on:
+  push:
+    branches: [main]
+  pull_request:
+    # `labeled` is not one of the default activity types; without naming them
+    # explicitly, adding the label would not trigger anything.
+    types: [opened, reopened, synchronize, labeled]
+
+concurrency:
+  group: bench-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  bench:
+    name: benchmark triggers
+    if: github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'perf')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+      # nits-review-core and nitsd pull in neither tauri nor the web bundle.
+      - run: NITS_BENCH_FILES=2000 cargo bench -p nits-review-core -p nitsd
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,15 +1,12 @@
 # Benchmarks are a regression *trigger*, not a correctness gate, and cost a
-# whole release-profile build of the workspace. They live in their own workflow
-# so that opting a PR in does not re-run every other CI job, and so `labeled`
-# can be an activity type here without making every label change re-run CI.
+# whole release-profile build of the workspace. They run on main only, never in
+# PR CI. `workflow_dispatch` is the escape hatch for checking a branch by hand
+# before merging it.
 name: bench
 on:
   push:
     branches: [main]
-  pull_request:
-    # `labeled` is not one of the default activity types; without naming them
-    # explicitly, adding the label would not trigger anything.
-    types: [opened, reopened, synchronize, labeled]
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
@@ -17,22 +14,10 @@ env:
 jobs:
   bench:
     name: benchmark triggers
-    # On `labeled`, gate on the label that was just added: otherwise, once a PR
-    # carries `perf`, adding any *other* label satisfies a whole-label-set check
-    # and starts another release build. Every other activity type checks the
-    # full set, since no single label is being added.
-    if: >-
-      github.event_name == 'push'
-      || (github.event.action == 'labeled' && github.event.label.name == 'perf')
-      || (github.event.action != 'labeled'
-          && contains(github.event.pull_request.labels.*.name, 'perf'))
-    # Job-level, not workflow-level: workflow concurrency is applied before a
-    # job's `if` is evaluated, so a run that is going to skip would still cancel
-    # a benchmark already in flight.
+    runs-on: ubuntu-latest
     concurrency:
       group: bench-${{ github.ref }}
       cancel-in-progress: true
-    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
       - uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,7 @@ jobs:
           cache: pnpm
           cache-dependency-path: ui/pnpm-lock.yaml
       - run: pnpm install --frozen-lockfile
+      - run: pnpm rescript format --check
       - run: pnpm rescript
       - run: pnpm test
       - run: pnpm vite build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,34 +29,50 @@ jobs:
           bash <(curl -sSfL https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
           ./actionlint -color
 
-  lint:
-    name: fmt, clippy, wasm
+  # Needs rustfmt and nothing else — no cache, no toolchain build, no bundle.
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: rustfmt
+      - run: cargo fmt --all --check
+
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+      # rust-cache adds the job id to the key by default, so each of these
+      # jobs gets its own cache without an explicit `key`.
+      - uses: Swatinem/rust-cache@v2
+        with:
+          # Caches are branch-scoped and PR branches are evicted quickly, so
+          # only main's cache is worth keeping; PRs restore from it.
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+      # Only clippy needs these: it is the job that compiles the tauri and
+      # web crates.
+      - uses: ./.github/actions/web-bundle
+      - run: cargo clippy --workspace --all-targets -- -D warnings
+
+  # nits-protocol and nits-client-core are sans-I/O and pull in neither tauri
+  # nor the embedded bundle, so this needs no web-bundle setup.
+  wasm:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: wasm32-unknown-unknown
-          components: clippy, rustfmt
-      # Per-job cache key: this job builds check artifacts, `test` builds dev
-      # with codegen and `bench` builds release. One shared cache would have
-      # them overwriting each other's profile on every run.
       - uses: Swatinem/rust-cache@v2
         with:
-          key: lint
-          # A single lint failure on main would otherwise leave main with no
-          # cache at all, and every PR branched from it then builds cold.
-          cache-on-failure: true
-          # Caches are branch-scoped and PR branches are evicted quickly, so
-          # only main's cache is worth keeping; PRs restore from it.
           save-if: ${{ github.ref == 'refs/heads/main' }}
-      - uses: ./.github/actions/web-bundle
-      - run: cargo fmt --all --check
-      - run: cargo clippy --workspace --all-targets -- -D warnings
       - run: cargo check -p nits-protocol -p nits-client-core --target wasm32-unknown-unknown
 
   test:
-    name: tests
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
@@ -64,34 +80,11 @@ jobs:
       - uses: taiki-e/install-action@nextest
       - uses: Swatinem/rust-cache@v2
         with:
-          key: test
-          # A single lint failure on main would otherwise leave main with no
-          # cache at all, and every PR branched from it then builds cold.
-          cache-on-failure: true
-          # Caches are branch-scoped and PR branches are evicted quickly, so
-          # only main's cache is worth keeping; PRs restore from it.
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - uses: ./.github/actions/web-bundle
       - run: cargo nextest run --workspace
       - name: fixtures are up to date
         run: cargo xtask fixtures && git diff --exit-code -- fixtures
-
-  # Benchmarks are a regression *trigger*, not a correctness gate, and cost a
-  # whole release-profile build of the workspace. Run them on main, and on a
-  # PR only when it is labelled `perf`.
-  bench:
-    name: benchmark triggers
-    if: github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'perf')
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v7
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
-        with:
-          key: bench
-          cache-on-failure: true
-          save-if: ${{ github.ref == 'refs/heads/main' }}
-      - run: NITS_BENCH_FILES=2000 cargo bench -p nits-review-core -p nitsd
 
   ui:
     runs-on: ubuntu-latest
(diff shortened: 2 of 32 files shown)


JonoPrest (commented):
Review by gpt-5.6-sol via Codex. 4 findings: 2 P1, 1 P2, 1 P3.

The broad split is working: the latest cold run completed in 4m16s versus the reported 8.5m sequential run, while scripts and UI finished in seconds. The action-version bumps also point at real tags and retain the inputs used here. Two workflow semantics need correction before this reliably does what the description promises, and the Rust lint group can be split more cleanly.

[P3 / nit] Granularity and cache notes (non-blocking)

- Keep shellcheck + actionlint together and keep the UI steps together; they share setup and finish quickly.
- Do not shard nextest yet: in the latest run compilation took 2m32s and test execution only about 28s, so naive shards would duplicate the dominant work.
- The fixture freshness check is the next plausible small split: `xtask` does not need Tauri libraries or the web bundle, and a nextest failure currently suppresses it. Against that, its current 14s benefits from the test job's artifacts, so measure the standalone version before keeping it.
- Main-only Rust cache writes are defensible here: the existing Linux cache is about 1.1 GiB and PR-scoped copies can evict useful main caches under GitHub's quota. The explicit `key: lint|test|bench` values are redundant, though, because rust-cache already includes `GITHUB_JOB` by default.
- After one successful `main` run warms the new keys, measure again. If tests remain the critical path, trial `CARGO_PROFILE_TEST_DEBUG=0`; move to sccache only if that measurement still justifies the extra machinery.

.github/workflows/ci.yml
```diff
@@ -39,31 +28,88 @@ jobs:
...
+          components: clippy, rustfmt
+      # Per-job cache key: this job builds check artifacts, `test` builds dev
+      # with codegen and `bench` builds release. One shared cache would have
+      # them overwriting each other's profile on every run.
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: lint
+          # A single lint failure on main would otherwise leave main with no
+          # cache at all, and every PR branched from it then builds cold.
+          cache-on-failure: true
```
JonoPrest: [P1 / major] Failed builds can permanently seed an incomplete exact-key cache

`rust-cache` saves after a miss/partial restore, but an exact hit is considered up to date and is never saved again; GitHub cache keys are immutable. If a new main key first encounters a format, web-build, or compile failure before dependencies finish, this setting publishes that partial cache. The fixed source usually does not change the toolchain/manifest key, so later runs restore the exact partial entry, compile the missing artifacts again, and cannot replace it. The same issue exists in `test` and `bench`.

The original actionlint failure no longer needs this workaround: `scripts` is independent, so successful Rust jobs still publish their own main caches even when `scripts` fails. Remove `cache-on-failure: true` from all three jobs and let only successful jobs establish a new exact key. The action's restore/save state flow is visible in [`restore.ts`](https://github.com/Swatinem/rust-cache/blob/6323deb102c322ba6fcbdcafc7e3dddab59af2b6/src/restore.ts) and [`save.ts`](https://github.com/Swatinem/rust-cache/blob/6323deb102c322ba6fcbdcafc7e3dddab59af2b6/src/save.ts).

.github/workflows/ci.yml
```diff
@@ -39,31 +28,88 @@ jobs:
         run: |
           bash <(curl -sSfL https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
           ./actionlint -color
+
+  lint:
+    name: fmt, clippy, wasm
```
JonoPrest: [P2 / minor] These three independent gates still suppress one another

A formatting failure ends the job before Clippy and wasm run, and a Clippy failure prevents the wasm check. They also have different setup requirements: formatting only needs rustfmt, and the two wasm-safe crates need neither the Tauri system libraries nor the web bundle. In the latest run, both waited about 39s for that bundle; wasm then waited another 1m33s for Clippy.

Split this into `fmt`, `clippy`, and `wasm` jobs, with `web-bundle` only in `clippy`. That gives independent results and much faster feedback without splitting the expensive nextest compile. rust-cache's automatic job-id keying will keep the Clippy and wasm target caches separate.

.github/workflows/ci.yml
```diff
@@ -39,31 +28,88 @@ jobs:
...
-        run: NITS_BENCH_FILES=2000 cargo bench -p nits-review-core -p nitsd
       - name: fixtures are up to date
         run: cargo xtask fixtures && git diff --exit-code -- fixtures
 
+  # Benchmarks are a regression *trigger*, not a correctness gate, and cost a
+  # whole release-profile build of the workspace. Run them on main, and on a
+  # PR only when it is labelled `perf`.
+  bench:
+    name: benchmark triggers
+    if: github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'perf')
```
JonoPrest: [P1 / major] Adding the `perf` label does not trigger this workflow

With bare `pull_request:`, GitHub only triggers on `opened`, `synchronize`, and `reopened`; `labeled` is not a default activity type. Adding `perf` to an existing PR therefore leaves this job skipped until some unrelated later push/reopen, so the advertised opt-in does not work when requested.

Add `labeled` while retaining the default activity types, or preferably move benchmarks to their own workflow triggered by main pushes plus PR `labeled`/`synchronize`/`reopened`. The separate workflow avoids rerunning every normal CI job just to activate benchmarks. See [GitHub's `pull_request` event activity types](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request).

JonoPrest: Fixed in 2217c03 — correct, and the PR description advertised behaviour that could not happen. `labeled` is not a default `pull_request` activity type, so the opt-in did nothing until an unrelated push.

I took your preferred option rather than just adding `labeled` to ci.yml: benchmarks now live in `.github/workflows/bench.yml` with `types: [opened, reopened, synchronize, labeled]`. Adding `labeled` to the main CI trigger would have made *every* label change re-run the whole suite, which is a worse trade than the extra file. I added `opened` as well, so a PR created with the label already applied is covered — GitHub does not fire `labeled` for labels present at creation.

.github/workflows/ci.yml
JonoPrest: Fixed in 2217c03 — you're right, and I checked the mechanism before removing it rather than taking it on trust. `save.ts` opens with `if (isCacheUpToDate()) { core.info('Cache up-to-date.'); return; }`, so an exact-key hit never re-saves, and GitHub keys are immutable — a partial cache written by a failed run is permanent until the toolchain/manifest key moves. That is strictly worse than the cold build it was meant to prevent.

And the justification is gone anyway: `scripts` is an independent job now, so the actionlint failure that started this can no longer stop the Rust jobs publishing their own main caches. Removed from all three. `save-if` stays.

.github/workflows/ci.yml
JonoPrest: Fixed in 2217c03 — split into `fmt`, `clippy` and `wasm`.

`fmt` takes rustfmt only, with no cache and no web-bundle, since `cargo fmt --check` compiles nothing. `wasm` skips the bundle too — I verified `nits-protocol` and `nits-client-core` contain no `include_dir!`, so neither embeds it. `clippy` keeps web-bundle as the only job that compiles the tauri and web crates.

So the two cheap gates no longer wait ~39s for a bundle they never use, and none of the three hides the others' results any more.

.github/workflows/ci.yml
JonoPrest: On the cache-key nit from the review body — taken in 2217c03. The explicit `key: lint|test|bench` values are gone; `config.ts` appends `GITHUB_JOB` when `add-job-id-key` is set, which is the default. The release workflow already demonstrates it: `v0-rust-plan-…` and `v0-rust-crates-io-…` are job-derived with no `key` input set. The new `fmt`/`clippy`/`wasm`/`test` jobs therefore get distinct caches for free.

On the rest of that note: agreed on not sharding nextest given 2m32s compile vs ~28s execution, and I've left the fixtures check inside `test` for exactly the reason you give — it currently benefits from that job's artifacts, so splitting it should follow a measurement, not precede one. Same for `CARGO_PROFILE_TEST_DEBUG=0`: worth trialling once main has warmed the new keys and there is a warm baseline to compare against, rather than now against a cold one.

.github/workflows/bench.yml
```diff
@@ -0,0 +1,33 @@
...
+  group: bench-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  bench:
+    name: benchmark triggers
+    if: github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'perf')
```
JonoPrest: [P2 / minor] Any later label restarts an already opted-in benchmark

Once a PR has `perf`, every subsequent `labeled` event satisfies this condition. For example, adding `perf` starts the release build; adding `ready-to-merge` a moment later starts another run, and the workflow-level concurrency group cancels the first. That wastes the expensive build and can repeatedly restart it while labels are being organized.

For `labeled`, gate specifically on `github.event.label.name == 'perf'`; for `synchronize`/`opened`/`reopened`, keep checking the full label set. Also move `concurrency` onto the `bench` job (or otherwise give irrelevant-label runs a different group), because workflow-level concurrency is applied before a skipped job can protect an in-progress benchmark.

JonoPrest (commented):
Follow-up review by gpt-5.6-sol via Codex against `2217c03`. The three original findings are correctly addressed. One new P2 edge case remains in the extracted benchmark workflow.

.github/workflows/ci.yml
JonoPrest: Verified in 2217c03 — fmt, Clippy, and wasm are independent, and only Clippy carries the web/Tauri setup.

.github/workflows/ci.yml
JonoPrest: Verified in 2217c03 — `cache-on-failure` is removed from lint/test/bench, so only successful main jobs establish exact keys.

.github/workflows/ci.yml
JonoPrest: Verified in 2217c03 — benchmarks moved to a separate workflow with the missing PR activity types.

.github/workflows/bench.yml
JonoPrest: Fixed in 526fd64 — both halves, and the concurrency one I would not have spotted.

`labeled` now gates on `github.event.label.name == 'perf'`, the label actually being added; `opened`/`reopened`/`synchronize` keep the whole-set check since no single label is in play for those. So adding `ready-to-merge` to an already-opted-in PR no longer starts a second release build.

Concurrency moved to the job. Your point that workflow-level concurrency is applied before the job's `if` is evaluated is the part that makes this a real bug rather than cosmetic: an irrelevant-label run would enter the group, cancel the in-flight benchmark, and then skip its own job — killing the build for nothing. At job level a skipped job never joins the group.

JonoPrest: Verified in 526fd64 — `labeled` now keys off the newly added label, and job-scoped concurrency leaves an active benchmark untouched when an irrelevant label produces a skipped job.

JonoPrest: Superseded in 7843473 — the repo owner has decided benchmarks should not run in PR CI at all, opt-in or otherwise.

So the `pull_request` trigger is gone entirely, and with it `labeled`, the label gating and the job's `if`. Both edge cases you raised are removed rather than fixed: there is no label event to mis-gate, and no run that starts only to skip. Job-level concurrency stays anyway, since back-to-back pushes to main should still cancel the older run — your reasoning about `if` being evaluated after the concurrency group is what made me keep it there rather than move it back up.

`workflow_dispatch` remains as the way to benchmark a branch by hand before merging.

JonoPrest (commented):
[P1 / major] The PR description now contradicts the final workflow

The description still calls `cache-on-failure: true` the substantive fix, says benchmarks run on PRs labelled `perf`, and says the three compiles use explicit cache keys. Those are now all false: failed-build saving was removed, benchmarks are main/manual-only, Rust checks are split more finely, and cache separation comes from automatic job-id keys. Because this PR is specifically documenting a CI/cache redesign, leaving the old rationale in the merge record would direct the next maintainer toward the exact failure mode fixed during review.

Please update the description to reflect the final design: independent scripts/fmt/Clippy/wasm/test jobs let successful jobs publish caches even when another gate fails; rust-cache's job-id keys separate profiles; only successful main jobs save; benchmarks run on main or manual dispatch.

JonoPrest:
Fixed — PR description and title rewritten to describe the final workflow, verified against the live body rather than the edit command's exit.

Corrected specifically:
- `cache-on-failure: true` is no longer presented as the fix. The description now says the *split* is the fix — `scripts` being an independent job is what makes a workflow-lint failure unable to stop the Rust jobs caching — and records why `cache-on-failure` was removed, so the next maintainer doesn't reintroduce it.
- Benchmarks are described as main-push or manual-dispatch only. The `perf` label opt-in and both edge cases you found are recorded as removed, not as behaviour.
- The explicit cache-key claim is gone; it now says separation comes from rust-cache's automatic `GITHUB_JOB` keying.
- Job table matches the actual `scripts`/`fmt`/`clippy`/`wasm`/`test`/`ui` split.

Worth flagging: my first attempt used `gh pr edit`, which failed with a GraphQL Projects-classic error while my shell still printed a success line. The body was unchanged and I nearly reported it done. Reissued through the REST endpoint and diffed the live body for both the removed and the added claims.

JonoPrest:
[P2 / minor] The description is almost aligned; two cache/job claims remain too strong

The `clippy` row calls it “the only job compiling tauri/web crates,” but `test` runs `cargo nextest --workspace` and therefore also installs `web-bundle`; the workflow reflects that at `test` line 88. Say that Clippy and tests need the bundle, while fmt/wasm do not.

Also, PR-scoped caches are not necessarily “evicted before [they are] reused”: subsequent runs of the same PR can restore them (this repository’s existing `refs/pull/1/merge` entries show later access). The main-only policy is still sensible with roughly 1.1 GiB Rust caches and GitHub’s shared quota; the accurate rationale is avoiding large, short-lived PR duplicates that can evict the broadly reusable main cache.

With those wording fixes, the body will match both GitHub cache scoping and the final workflow.

JonoPrest:
Fixed — both claims corrected, verified against the live body.

The `clippy` row no longer claims exclusivity: `clippy` and `test` both need `web-bundle` (`test` runs `nextest --workspace`, so it compiles the tauri and web crates too), while `fmt` and `wasm` do not. The table now says that.

On PR-scoped caches: you're right and I checked rather than just softening the wording — this repo currently has 3 `refs/pull/*` cache entries whose `last_accessed_at` postdates their creation, so they demonstrably do get restored by later runs of the same PR. The rationale now says what actually justifies main-only writes: at ~1.1 GiB each they are short-lived duplicates that can evict the broadly reusable main cache under GitHub's shared per-repo quota — not that they go unused.

JonoPrest (commented):
LGTM from gpt-5.6-sol via Codex — 7 findings addressed, 0 withdrawn, 0 nits left to the author's discretion. All six PR CI checks are green at `7843473`; the cache policy, job boundaries, benchmark scope, and PR description now agree.

JonoPrest:
Closing summary — 4 rounds, 7 findings, all fixed, none pushed back on.

**Code**
- `labeled` is not a default `pull_request` activity type, so the `perf` opt-in never triggered. Benchmarks moved to their own workflow, then to main-push/manual-dispatch only once Jono decided they should not run on PRs at all — which removed the two follow-on edge cases (a later label restarting an opted-in run; workflow-level concurrency cancelling an in-flight benchmark before the job's `if` was evaluated) rather than fixing them.
- `cache-on-failure: true` was removed. Verified in `save.ts` that an exact-key hit is never re-saved, so a run failing midway would have seeded a permanent partial cache. The split makes it unnecessary: `scripts` is independent, so a workflow-lint failure cannot stop the Rust jobs caching.
- Lint split into `fmt` / `clippy` / `wasm` so they stop hiding each other, with `web-bundle` only where it is compiled.
- Explicit cache keys dropped as redundant — rust-cache appends `GITHUB_JOB` by default.

**Description** (rounds 3-4) — rewritten to match the built design rather than the proposed one, then corrected twice more: `clippy` is not the only job needing `web-bundle` (`test` runs `--workspace`), and PR-scoped caches are not evicted before reuse (this repo has 3 `refs/pull/*` entries whose `last_accessed_at` postdates creation). Both were checked against real data, not reworded.

**Deferred, with reasons on the thread:** splitting the fixtures check out of `test`, and `CARGO_PROFILE_TEST_DEBUG=0`. Both should follow a warm-cache measurement rather than precede one — there is no warm baseline until this merges and main saves under the new job-id keys.

Measured effect: 8.5m sequential cold → 4m16s split cold.

PR #7, opened 2026-09-02
One binary: fold the daemon and the MCP server into `nits`

JonoPrest:
Installing Nits meant installing **two** binaries — `nits` plus `nitsd`, or `nits-mcp` plus `nitsd` — because the clients start the daemon by looking for `nitsd` next to themselves. Every channel worked around that differently (archives/formula/AUR bundled it, deb/rpm declared `Depends: nitsd`, cargo could do neither so the README said `cargo install nits nitsd`), and nothing kept the two in step: a channel that updated one and not the other leaves a client and a daemon on different protocol versions, and the handshake fails.

`nitsd` and `nits-mcp` lose their `[[bin]]` sections and become libraries only. `nits` already linked both, so this is mostly deletion.

- **Daemon** is `nits daemon serve`; **MCP server** is `nits mcp`.
- `launch::nits_binary` starts a daemon by **re-executing the running `nits`** (`$NITS_BIN` overrides; other embedders look next to themselves, then `PATH`). Client and daemon are the same build by construction.
- `nitsd::serve` holds what `nitsd`'s `main` did; `DaemonSpec` gained `argv_prefix`, defaulting to `daemon serve`.
- ssh contexts run `ssh <host> nits daemon stdio`. `Context::Ssh.nitsd` → `.bin`, serde-aliased so existing config files still load; `--no-autostart` replaced `--stdio-if-running`.
- `daemon serve|stdio` take `--ws-listen`, not `--ws` — the global `--ws` is the client side, a URL to dial.
- Packaging collapses to one artifact per channel: no bundling, no `Depends`, `cargo install nits` is the whole instruction. `Releasable` has one variant; the workflow's package choice is `[nits]`.
- Daemon lifecycle tests moved to `nits-cli`, where the binary they spawn now lives.

`nits-web` and `nits-desktop` are untouched — a dev bridge and a GUI app, neither installed from a package.

## A bug this caught

The first cut named the serve flag `--ws-listen` but left the field `ws`. clap keys arguments by *field name* and only notices a clash when the value is read, so `nits daemon stdio --ws-listen …` panicked with `Mismatch between definition and access of 'ws'` — at run time, in front of a user. Neither `--help` nor `Command::debug_assert()` catches it; `the_daemon_subcommands_parse_their_flags` does, and was checked against the reintroduced bug.

## Verification

- `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, `cargo test --workspace` — all green.
- `shellcheck scripts/*.sh`, `cargo xtask fixtures` no-op, `cargo xtask release-plan --package nits` ships `["nits"]`.
- End-to-end on the built binary: autostart via `workspace list` (spawns `nits daemon serve --data-dir … --socket …`), `nits mcp` `initialize` returns `Connected to nitsd 0.1.0`, `nits daemon stdio --no-autostart` exits 3, `--ws-listen 127.0.0.1:7701` really listens, and `nits .` serves the web UI (HTTP 200) with the deep-linked review.

https://claude.ai/code/session_01TgVYCVE8YPrc1o3t1GRupG


crates/nits-config/src/lib.rs
```diff
@@ -65,14 +65,16 @@ pub enum Context {
...
+    /// from `~/.ssh/config`.
     Ssh {
         host: String,
-        /// The remote `nitsd` binary. Default: `nitsd` on the remote PATH.
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        nitsd: Option<String>,
+        /// The remote `nits` binary. Default: `nits` on the remote PATH.
+        /// The `nitsd` alias reads configs written when the daemon was a
+        /// second executable.
+        #[serde(default, alias = "nitsd", skip_serializing_if = "Option::is_none")]
```
JonoPrest: [P2 / minor] The legacy alias preserves parsing but silently changes what gets executed

An existing config such as `nitsd = "/opt/bin/nitsd"` now fills `bin`, after which `ssh_command` runs `/opt/bin/nitsd daemon stdio`. The legacy binary expects `--stdio` (and may be removed during the upgrade), so status is reported as stopped and connections fail even though the config "still loads." Please keep the legacy spelling distinguishable long enough to emit a targeted migration error, or otherwise migrate it to a valid `nits` path, and add a test for the old serialized form.

JonoPrest (commented):
Review by gpt-5.6-sol via Codex. 2 findings: 0 P1, 2 P2, 0 P3.

The one-binary routing, lifecycle tests, and package builders look coherent; the targeted CLI/lifecycle tests, xtask tests, and clippy all pass locally. Two migration/release surfaces still need attention.

[P2 / minor] Update the README that the `nits` crate actually publishes

`crates/nits-cli/Cargo.toml` still declares its crate-local `README.md`, and `cargo package -p nits --list` confirms that file is shipped. It currently instructs users to run `cargo install nits nitsd` and says agents attach through `nits-mcp`; after this PR both dependency crates have no binary target, so that install command fails and the crates.io README contradicts the one-binary release. Please update that README to `cargo install nits` / `nits mcp`; the user-facing `AddSsh` help in `main.rs` also still says `ssh HOST nitsd --stdio`.

crates/nits-config/src/lib.rs
JonoPrest: Fixed in 51d7072 — you're right, and the failure mode was worse than "still loads": `/opt/bin/nitsd daemon stdio` is not a command the old daemon parses, so the host reported as unreachable for a reason invisible to the user.

The alias is gone. `nitsd` now deserializes into its own `legacy_nitsd` field, deliberately *not* onto `bin`, and `contexts::connect` refuses the context with the edit to make. `connect` is the single choke point — `status`, `start` and `stop` all route through it — so one check covers every path. `daemon status` never fails the process, so the guidance lands in the row itself:

```
box  ssh build-box  unreachable: this context names a `nitsd` binary on build-box
(/opt/bin/nitsd), which cannot serve: the daemon is now `nits daemon serve`.
Replace `nitsd = "..."` with `bin = "nits"` (or the path to `nits` on that host),
or drop the line to use `nits` from the remote PATH
```

I did not auto-migrate `/opt/bin/nitsd` to `/opt/bin/nits`: it's a guess about someone's remote filesystem, and guessing wrong produces the same invisible failure this fix exists to remove.

`nits context add-ssh` also lost its `--nitsd` alias, which would have written the same broken context.

Tests: `the_pre_one_binary_nitsd_key_is_recognised_not_silently_reused` in `nits-config` asserts the old serialized form loads, lands in `legacy_nitsd`, and leaves `bin` as `None` — plus the migrated form for contrast. `a_legacy_ssh_context_says_how_to_migrate_it` in `nits-cli` drives the real binary through both `daemon status` and a failing command.

JonoPrest:
Fixed in 51d7072 — thanks, I'd missed that `crates/nits-cli/README.md` is a separate file from the repo root one and is what crates.io publishes. `cargo package -p nits --list` confirms it ships.

It now says `cargo install nits` and `nits mcp`, and explains that the daemon is the same binary rather than a second crate to install. Also updated in the same commit:

- `context add-ssh`'s help, which still said `ssh HOST nitsd --stdio` → `ssh HOST nits daemon stdio`.
- The `--bin` flag's help, noting there is deliberately no `--nitsd` alias.
- `nits-cli`'s module doc, which described the crate as "command-line client for `nitsd`".

The second finding (the legacy `nitsd` config key) is answered on its own thread.

Gates re-run on 51d7072: `cargo fmt --check`, `clippy --workspace --all-targets -D warnings`, `cargo test --workspace`, `shellcheck scripts/*.sh`, `cargo xtask fixtures` no-op.

crates/nits-config/src/lib.rs
JonoPrest: [P2 / minor] The behavior is explicit now, but the domain still represents the invalid two-key state

`Context::Ssh` now carries both `bin: Option<String>` and `legacy_nitsd: Option<String>`, so a config containing both keys deserializes successfully and the core has to choose one. That is the exact `Option + Option` shape AGENTS.md rules out ("Enums over flags ... Never Option<a> + Option<b>"). Please parse the two wire spellings once at the config boundary into one exhaustive domain value such as `Default | Nits(path) | LegacyNitsd(path)`, rejecting both keys together, and have `contexts::connect` match that enum.

The migration diagnostic and its tests otherwise behave correctly; the incremental test suite passes locally.

JonoPrest:
Verified in 51d7072 — the crate-local README now ships the one-binary install/MCP commands, and the stale SSH help/module documentation is corrected. This finding is resolved.

JonoPrest:
**Round 1 closed.** Both P2 findings fixed in 51d7072; all six checks green on that commit (fmt, clippy, test, wasm, ui, scripts).

| Finding | Disposition |
| --- | --- |
| Legacy `nitsd` config key silently changes what gets executed | Fixed — old key deserializes into its own `legacy_nitsd` field, `contexts::connect` refuses the context with the exact edit to make; two tests |
| Published crate README still says `cargo install nits nitsd` | Fixed — `crates/nits-cli/README.md` rewritten, plus `add-ssh` help and the CLI module doc |

Nothing deferred and nothing pushed back on. The one judgement call worth flagging: I did **not** auto-migrate `/opt/bin/nitsd` to a sibling `/opt/bin/nits`. It's a guess about a remote filesystem, and a wrong guess reproduces the invisible failure the fix exists to remove — so the user is told what to write instead.

crates/nits-config/src/lib.rs
JonoPrest: Fixed in cbe4322 — the two spellings are parsed into one domain value at the boundary. `ContextWire` holds the optional `bin`/`nitsd` keys and `serde(try_from/into)` converts it once into `RemoteBin = Default | Nits(path) | Legacy(path)`; both keys together is `ContextParseError::BothBinAndNitsd`, so the file is refused with the edit to make rather than one silently winning. `contexts::connect` matches the enum exhaustively and `RemoteBin::nits()` is `None` for `Legacy`, so nothing downstream can run it by accident. Tests cover the old spelling parsing to `Legacy` and round-tripping back to `nitsd = "..."` (an old client still reads what we write), both keys as a load error, `bin` alone, and neither key.

Also rebased onto main (which now has #13).
— claude-opus-5 via Claude Code (author)

JonoPrest: Verified in `cbe4322`: `ContextWire` contains the two optional wire spellings only at deserialization, `TryFrom` rejects both together, and `Context::Ssh` carries the exhaustive `RemoteBin` domain value. Serialization preserves the legacy spelling, runtime connection handling cannot execute it, the focused config/CLI/daemon tests pass, targeted clippy is clean, and all CI jobs are green. Resolved.

JonoPrest (commented):
Re-reviewed `cbe4322`. The remaining P2 is fixed and resolved: the wire-only dual spelling is parsed into one exhaustive domain enum, the invalid both-keys form is rejected, focused validation passes, and CI is green. No P1/P2 findings remain.

PR #23, opened 2026-09-03
Detect the default review base beside the repository

### LINKED ISSUE #22: Default review base should be the branch this branch was cut from, not a hardcoded `main`
## What happens

`nits .` in a repo whose default branch is `master`, with no `origin`:

```console
$ nits --headless .
Error: daemon: Internal { message: "cannot resolve refs/heads/main: no such revision" }
```

`main` was never in the repo. It came from a hardcoded fallback.

## Why

`crates/nits-cli/src/main.rs`:

```rust
fn default_branch(root: &Path) -> String {
    // git symbolic-ref --short refs/remotes/origin/HEAD, minus the "origin/" prefix
    ...
    .unwrap_or_else(|| "main".into())
}
```

`refs/remotes/origin/HEAD` is set by `git clone` and by nothing else. It is absent in a `git init` repo, absent after adding a remote by hand, absent on a `--single-branch` clone, and can be left dangling by pruning fetches. When it is absent, nits guesses `main` and review creation fails against a `master` repo.

Verified — in a `git init` repo with `master`, `feature-1`, `feature-2`:

```console
$ git symbolic-ref --short refs/remotes/origin/HEAD
fatal: ref refs/remotes/origin/HEAD is not a symbolic ref
```

Two more places compound it:

1. **Remote contexts skip detection entirely.** In `directory_review`, the non-local arm is literally `(path.to_path_buf(), "main".to_owned())` — no probing at all, because the CLI cannot shell out to git on the other machine. So `nits -c hetzner <path>` always guesses `main`.
2. **The first choice is `RefSpec::Upstream`**, with `default_branch` only as the fallback when that fails. For a branch that has been pushed, `@{upstream}` is *the same branch* — base and head coincide and the review shows only unpushed work. That is rarely what "review my branch" means.

## What it should do

Default base = **the branch this branch was created from**. That is the right answer for the ordinary case *and* for stacked branches, where the parent is another feature branch rather than the trunk.

Git records this imperfectly, so it needs a small ladder rather than one command:

- **Reflog.** `git reflog show <branch>` ends with `branch: Created from <X>`. Reliable when the branch was cut from a named branch:
  ```
  8e742ed feature-3@{0}: branch: Created from master
  ```
  But `git checkout -b` from the current branch records `Created from HEAD`, which names nothing:
  ```
  acc0396 feature-2@{1}: branch: Created from HEAD
  ```
  Reflogs also expire (90 days) and don't survive a fresh clone.

- **Closest-ancestor branch.** For each other branch `B`, take `git merge-base B HEAD` and count `git rev-list --count <mb>..HEAD`; the smallest count is the branch you most recently diverged from. On the stack above (`master` → `feature-1` → `feature-2`, HEAD on `feature-2`):
  ```
  master:    merge-base=8e742ed  commits-ahead=2
  feature-1: merge-base=acc0396  commits-ahead=1     ← correct parent
  ```
  This is the one that actually handles stacked branches, and it needs no reflog and no remote.

- **Trunk detection, for when HEAD *is* the trunk or nothing else matches.** Should not assume a name: try `refs/remotes/origin/HEAD`, then `git config init.defaultBranch`, then which of `main`/`master`/`trunk`/`develop` actually exists (`git rev-parse --verify`), then — if the repo has exactly one branch — that branch. Never fall back to a name that isn't in the repo: if nothing resolves, say so instead of failing later with `cannot resolve refs/heads/main`.

Ambiguity is real (several branches can share a merge-base) so ties want a deterministic rule and the chosen base should be stated in the "review created" line, not silently assumed.

## Where it should live

This detection currently sits in the CLI and shells out to `git`, which is exactly why remote contexts get `"main"` hardcoded. The daemon already owns a git engine (`nits-review-core`) and already resolves `RefSpec`s against the repo on *its* machine. Putting base detection there — a request the CLI asks for, rather than a computation it performs — fixes local and remote in one move and stops the two paths from drifting.

## Acceptance

- `nits .` works in a `master` repo with no remote, and in a repo whose trunk is neither `main` nor `master`.
- A branch cut from another feature branch defaults to that branch, not the trunk.
- Works identically over `-c <remote>`.
- When no base can be determined, the error names the repo and what was tried — never a ref the repo does not have.

JonoPrest:
Fixes #22

## Summary
- move default-base discovery into the daemon so local and remote contexts agree
- choose a named reflog parent, closest ancestor branch, or detected trunk through a deterministic typed request
- report the chosen base when creating/opening a directory review and return an actionable error when none exists
- cover master-only repos, nonstandard trunks, stacked branches, descendant rejection, CLI output, and SSH contexts

## Validation
- `cargo test -p nits-review-core --test git`
- `cargo test -p nits --test cli directory_review_uses_and_reports_the_daemons_detected_base -- --exact`
- `cargo test -p nits --test cli contexts_and_daemon_lifecycle -- --exact`
- protocol/UI fixture and clippy checks in progress locally; CI is authoritative


crates/nits-review-core/src/git.rs
```diff
@@ -117,8 +140,177 @@ impl Repo {
...
+            Ok(head) => head,
+            Err(GitError::Resolve { .. }) => return Ok(None),
+            Err(error) => return Err(error),
+        };
+        let mut ranked = Vec::new();
+        for branch in branches {
+            if current == Some(branch) {
+                continue;
+            }
+            let branch_tip = self.rev_parse_commit(branch.as_str())?;
```
JonoPrest: [P2 / minor] Resolve candidate branches through `refs/heads/`, not Git's short-name DWIM

`local_branches` guarantees this value names a local branch, but `rev_parse_commit(branch.as_str())` resolves the short spelling using Git's disambiguation rules. If a tag has the same name, Git can resolve the tag here; the ranking then measures that tag commit but returns `RefSpec::Branch`, which later resolves the different branch commit. For example, with a divergent `alpha` branch, an `alpha` tag at HEAD's merge base, and the actual ancestor branch `zbase`, this code can rank `alpha` as pointing at the merge base and then create the review against divergent `refs/heads/alpha`. Please use a fully qualified `refs/heads/{name}` consistently for `rev_parse_commit` and the merge-base probes, and add a branch/tag collision test.

JonoPrest (commented):
Review by gpt-5.6-sol via Codex. 1 finding: 0 P1, 1 P2, 0 P3.

The daemon-side design and coverage look sound overall, and CI is green. One Git ref-resolution edge case can make the detector return a branch different from the revision it ranked.

crates/nits-review-core/src/git.rs
JonoPrest: Fixed in 1733e4d — candidate branches are now discovered and ranked exclusively through fully qualified `refs/heads/*` names. Added a branch/tag collision regression that proves the selected branch is the revision that was ranked.

JonoPrest: Verified in 1733e4d — candidate discovery, revision parsing, and both merge-base probes now use fully qualified local-branch refs; the collision regression passes locally. Resolved.

JonoPrest (commented):
LGTM from gpt-5.6-sol via Codex — 1 finding addressed, 0 withdrawn, 0 nits left to the author's discretion. Focused regression and CLI/core tests pass; all required CI checks are green.

JonoPrest:
Review loop complete: fixed the branch/tag ambiguity in `1733e4d` by qualifying every candidate as `refs/heads/*` and adding a collision regression. No feedback was deferred or pushed back on; the reviewer verified the fix, all six CI checks passed, and PR #23 was squash-merged as `8b62a0d`.

PR #24, opened 2026-09-03
Let UI hosts connect through remote contexts

### LINKED ISSUE #21: Web UI can't open a remote context: `nits -c <ssh|ws>` bails even when every CLI command works
## What happened

```console
$ nits --context hetzner
Error: the web UI needs a local context so far (remote viewing: PLAN 4.6); `--headless` works on remote contexts
```

The daemon on the remote is reachable and healthy — `nits -c hetzner workspace list`, `review list`, `files`, `diff` and `comment` all work over the same context. Only the thing you actually want to look at, the browser UI, is unavailable. Creating the review remotely works (`nits --headless . review create --base master --head jono-false-positive` → `01M1KP13X8FTBD17BAVGQJTVXK`), and then there is no way to view it locally.

## It is not only ssh

A WebSocket context with a live daemon behind it fails identically:

```console
$ nits -c hetzner-ws workspace list      # works — the daemon answers over ws
$ nits -c hetzner-ws
Error: the web UI needs a local context so far ...
```

So today there is **no** context type that can drive the web UI except an implicit/explicit local one. Reproduced without any remote machine, using a stand-in `ssh` that runs `nits daemon stdio` against a second data dir (the pattern `crates/nits-cli/tests/cli.rs` already uses).

## Why

`crates/nits-cli/src/main.rs` (`open_ui`) needs a **unix socket path** to hand to the host, and a remote context has none:

```rust
let socket = match ctx {
    Context::Local { data_dir, socket } => nitsd::contexts::local_spec(...)?.socket,
    Context::Ssh { .. } | Context::Ws { .. } => anyhow::bail!("the web UI needs a local context so far ..."),
};
```

The constraint comes from `nits-client-host`: `HostConfig.socket: PathBuf`, and its dialer is hardcoded:

```rust
async fn connect(socket: PathBuf, incoming: ...) {
    match tokio::net::UnixStream::connect(&socket).await { ... }
}
```

Everything above that is already transport-agnostic, which is what makes this look tractable:

- `serve_stream` in `nits-client-host` is already generic over `AsyncRead + AsyncWrite`, so the ssh-stdio case needs no new plumbing — just a different way to get the stream.
- `nitsd::transport` already has `FrameRead`/`FrameWrite` with both a `byte_stream` impl (unix, ssh stdio) and a `web_socket` impl. Generalising `HostConfig` to those traits covers all three transports uniformly, rather than special-casing ws.
- `nitsd::contexts::connect` already builds exactly these streams for the CLI's `Client` — ssh via `Stdio::piped()` on `ssh <host> nits daemon stdio`, ws via `connect_ws`. The host wants the same thing one layer down.

Sketch: replace `HostConfig.socket: PathBuf` with the `nits_config::Context` (or a small `Endpoint` enum), and have the host's reconnect path go through a context-aware dialer. The reconnect loop matters — it currently redials the socket on `Disconnected`, and an ssh child process has a different liveness story than a unix socket.

## Same gap in the desktop app

`crates/nits-client-tauri/src/lib.rs` also calls `nitsd::contexts::local_spec` + `ensure_daemon`, so `--ui desktop` is local-only for the same reason. Whatever shape the fix takes should be shared, not duplicated per client.

## Workaround today

Run the UI **on** the remote and forward its port:

```console
$ ssh hetzner -L 9777:127.0.0.1:9777
remote$ nits --port 9777 .
```

`nits` binds loopback only (`SocketAddr::from((Ipv4Addr::LOCALHOST, cli.port))`) and there is no flag to bind elsewhere, so the forward is mandatory — which is fine, and arguably the right default, but worth knowing.

## Notes

- `PLAN.md` 4.6 lists "remote connection UX (ssh target picker)", which is the *UX* on top of this. The transport work underneath it is the actual blocker and isn't written down anywhere.
- The error message is honest and points at the plan item, which is good. It could also name the port-forward workaround.
- Unrelated but visible in the report: ssh's own stderr is inherited (deliberately, so auth prompts and host-key questions work), so a user's `LocalForward` lines in `~/.ssh/config` produce `bind [127.0.0.1]:8000: Address already in use` noise above the nits error. Nothing to fix, but it makes the failure look worse than it is.

JonoPrest:
Fixes #21

## Summary
- resolve Local, SSH, and WebSocket contexts into a typed daemon endpoint with an explicit start policy
- share one framed dial path between CLI/MCP clients and client-core hosts while preserving core-owned handshakes and reconnect gap replay
- own and reap each SSH stdio child with its transport
- enable named SSH/WS contexts in the browser bridge, standalone web binary, and Tauri host
- cover native WebSocket host reconnects and end-to-end named SSH/WS browser sessions

## Validation
- `cargo test -p nits-review-core -p nitsd -p nits-client-host -p nits-client-web -p nits-mcp -p nits`
- `cargo clippy -p nits-review-core -p nitsd -p nits-client-host -p nits-client-web -p nits-mcp -p nits --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

`nits-client-tauri` cannot compile in this container because the system `glib-2.0.pc` development package is absent; CI has the required WebKit/GLib dependencies and is authoritative for that crate.


crates/nits-cli/src/main.rs
```diff
@@ -426,12 +426,20 @@ fn identity(cli: &Cli) -> Identity {
...
-    let client = contexts::connect(ctx, identity(cli), !cli.no_autostart)
+    let client = contexts::connect(ctx, identity(cli), start_policy(cli.no_autostart))
         .await
         .with_context(|| format!("connecting to {}", ctx.describe()))?;
     Ok(Ops::new(client))
 }
 
+fn start_policy(no_autostart: bool) -> contexts::StartPolicy {
+    if no_autostart {
+        contexts::StartPolicy::RequireRunning
```
JonoPrest: I think I prefer just having --start-policy as a flag with these 2 options rather than boolean

JonoPrest: Fixed in 55576ae — replaced the boolean with an exhaustive `--start-policy <start-if-needed|require-running>` CLI enum, threaded the typed policy through UI/MCP/stdio paths, and updated the SSH invocation, docs, and parser/lifecycle tests.

JonoPrest (commented):
Review by gpt-5.6-sol via Codex. 2 findings: 0 P1, 2 P2, 0 P3.

The typed endpoint and shared framed dialer are a solid base, but two host-boundary cases still select the wrong behavior: desktop launch loses some CLI endpoint/lifecycle choices, and native remote hosts never turn on their persistent content cache.

crates/nits-cli/src/main.rs
```diff
@@ -720,26 +738,20 @@ async fn open_ui(cli: &Cli, ctx: &Context, ops: &mut Ops) -> anyhow::Result<()>
             if let Some(c) = &cli.context {
                 child.arg(c);
             }
+            if let Some(config) = &cli.config {
+                child.env("NITS_CONFIG", config);
```
JonoPrest: [P2 / minor] Preserve the resolved endpoint when launching the desktop UI

Only forwarding `cli.context` and `NITS_CONFIG` means an ad-hoc endpoint is discarded. For example, `nits --ws ws://remote:7677 --ui desktop` first connects to that WebSocket, then spawns `nits-desktop` with no context argument, so the app silently opens the implicit local daemon instead. The selected `--start-policy require-running` is likewise lost because `endpoint_for` always constructs `StartIfNeeded`. Please pass the already-resolved endpoint/lifecycle choice across the process boundary (or give `nits-desktop` equivalent typed arguments) and cover the ad-hoc WebSocket case.

crates/nits-client-host/src/lib.rs
```diff
@@ -428,17 +430,17 @@ where
...
     identity: Identity,
     id_seed: IdSeed,
     kv: KvConfig,
 ) -> HostConfig {
     HostConfig {
-        socket: socket.to_path_buf(),
+        endpoint,
         kv,
         identity,
         cache: CacheConfig::default(),
```
JonoPrest: [P2 / minor] Enable the disk tier for native remote endpoints

`CacheConfig::default()` sets `disk` to `Disabled`, and no production caller overrides it. Consequently a Tauri client using SSH/WebSocket still takes the no-disk `OpenReview` flow and refetches trees/renders after every app restart even though it has a Redb KV. That contradicts ARCHITECTURE §5.1’s remote-client contract and is particularly visible over SSH. Please choose `DiskTier::Enabled` (with the documented budget) for persistent SSH/WebSocket hosts while keeping local Unix and memory-only bridge sessions disabled, with a config-level test for both endpoint classes.

crates/nits-cli/src/main.rs
JonoPrest: Fixed in ee3e07e — `nits --ui desktop` now passes typed endpoint-source and start-policy arguments to `nits-desktop`; the desktop parses them into `EndpointSource`/`EndpointOptions` before resolving. Added coverage for the ad-hoc WebSocket URL and `RequireRunning` lifecycle choice.

crates/nits-client-host/src/lib.rs
JonoPrest: Fixed in ee3e07e — native hosts now enable the documented 2 GiB disk tier only for SSH/WebSocket endpoints backed by persistent Redb KV. Local endpoints and memory-only browser bridges remain disk-disabled, with config-level tests for all three cases.

crates/nits-client-tauri/src/main.rs
```diff
@@ -1,12 +1,83 @@
...
+use nitsd::contexts::StartPolicy;
+
+#[derive(Debug, Parser)]
+#[command(name = "nits-desktop", version, about)]
+struct Args {
+    /// Named context from the config file; the default context when absent.
+    #[arg(value_name = "CONTEXT", conflicts_with_all = ["socket", "ws", "data_dir"])]
+    context: Option<String>,
+    #[arg(long, conflicts_with_all = ["socket", "ws", "data_dir"])]
+    config: Option<PathBuf>,
```
JonoPrest, line 20: [P2 / minor] Keep honoring NITS_CONFIG in direct desktop launches

Before this refactor, endpoint_for read NITS_CONFIG when resolving a named context. The new parser only populates config from --config, so NITS_CONFIG=/tmp/team.toml nits-desktop now silently loads the default config instead (the parent nits launcher happens to translate its own env value into --config, but direct desktop launches remain documented/supported). Please add the same env binding used by the other CLI boundaries and cover parsing the environment-backed config path.

JonoPrest (commented):
Follow-up review of ee3e07e by gpt-5.6-sol via Codex. The two original findings are addressed, but the new desktop argument boundary drops one existing config source.

crates/nits-cli/src/main.rs
JonoPrest: Verified in ee3e07e: the launcher now forwards the selected endpoint source and start policy as typed desktop arguments, with coverage for the ad-hoc WebSocket case.

crates/nits-client-host/src/lib.rs
JonoPrest: Verified in ee3e07e: persistent SSH/WebSocket hosts receive the documented 2 GiB disk tier, while local and memory-only hosts remain disabled, with focused tests.

crates/nits-client-tauri/src/main.rs
JonoPrest: Fixed in 1e069ce (CI-safe test follow-up in 1e27b63) — the desktop `--config` argument now binds `NITS_CONFIG`, matching the main CLI boundary. The regression verifies the clap environment binding and exercises that typed config path through `EndpointSource`.

JonoPrest: Verified in 1e069ce: the desktop parser now binds NITS_CONFIG to the typed --config field, and the binary-level parser test covers the environment-backed path. Resolved.

JonoPrest:
All six CI checks are green at 1e27b63. The typed start-policy preference is implemented in 55576ae, all automated findings have been verified, and the later CI-safe test follow-up preserves the reviewed behavior. Ready for final confirmation.\n\n— gpt-5.6-sol via Codex (author)

JonoPrest (commented):
LGTM from gpt-5.6-sol via Codex — independent final review at `1e27b63` found no new P1/P2 issues. The typed endpoint/framing boundary covers Local, SSH, and WebSocket transports; SSH children are owned by the transport; reconnect creates a fresh dial; desktop arguments preserve ad-hoc and named endpoint sources plus lifecycle policy; and persistent remote hosts select the documented disk tier. The three earlier findings are addressed and verified, and all six CI checks are green.

JonoPrest:
Review loop complete at 1e27b63: the start policy is now an exhaustive typed CLI value; desktop launches preserve the resolved endpoint, lifecycle policy, and NITS_CONFIG source; persistent remote hosts receive the documented disk cache; and the CI-safe parser regression covers the config boundary. All six required checks are green, all findings have replies and pushed fixes, and the final independent re-review is LGTM.\n\n— gpt-5.6-sol via Codex (author)

PR #26, opened 2026-09-03
Give every browser WebSocket its own client session

### LINKED ISSUE #20: web: a session per connection, so tabs can hold different reviews
`nits-client-web` spawns one `nits_client_host` and fans its patches out to every socket:

```rust
let (handle, mut patches) = nits_client_host::spawn(config, shutdown.clone())?;
let (patches_tx, _) = broadcast::channel::<String>(256);
```

So every browser attached to the bridge is a mirror of one session: same focus, same open review, same cursor. Two tabs on two different reviews is not a missing switch — there is nowhere to put the second session.

The daemon is already built for many clients (per-client ids, independent subscriptions with gap replay, mutations tagged by client sequence), and the TUI and Tauri clients each own a core in their own process, so this is a limitation of the web bridge alone.

## What it needs

- A `ClientCore` per connection (or per session id, so a reload can rejoin), each with its own daemon connection.
- Patches routed to the connection that owns them rather than broadcast.
- A decision on what is per-session and what is per-user: focus and open review are obviously per-tab, while layout and whitespace prefs read as user-wide.
- Memory: a content cache per core. The daemon's render cache is shared, so it is mostly duplicated hot chunks, but the budget wants a look.

## Why it matters beyond the feature

It is also what makes the keyboard copy path exact in the browser. A clipboard write needs the gesture that asked for it, so the shell acts before the core answers and correlates afterwards against `ViewModel::last_key` (#17). With one core behind every socket, another tab's key advances that same count and can release a deferred copy early. A reply on the connection that sent the key removes the ambiguity entirely.

Filed from review on #17.

JonoPrest:
Stacked on #24. Review only 48cf170 here.

Fixes #20

## Summary
- create one ClientCore, typed client identity, host task, and daemon connection per browser WebSocket
- share only a once-opened KV store so preferences persist without sharing focus, review, cursor, or key state
- route patches only to their owning socket and cancel the complete framed connection on tab/server shutdown
- reset browser session state on reconnect and ignore events from stale sockets
- document the per-tab cache budget and shared-versus-session state boundary

## Validation
- cargo test -p nits-client-host -p nits-client-web -p nits
- cargo clippy -p nits-client-host -p nits-client-web -p nits --all-targets -- -D warnings
- pnpm --dir ui rescript
- pnpm --dir ui test (537 tests)
- pnpm --dir ui vite build
- cargo fmt --all -- --check
- git diff --check

Full-workspace desktop compilation remains unavailable locally because glib-2.0.pc is not installed; CI is authoritative for that crate.


crates/nits-client-host/src/lib.rs
```diff
@@ -401,32 +518,31 @@ where
...
+        tokio::select! {
+            outbound = out_rx.recv() => match outbound {
+                Some(msg) => {
+                    if transport::send_msg(&mut write, &Envelope::current(msg)).await.is_err() {
+                        break;
+                    }
+                }
+                None => break,
+            },
+            inbound = transport::recv_msg::<_, ServerMsg>(&mut read) => match inbound {
```
JonoPrest: [P1 / major] Do not cancel an in-progress byte-frame read

This `select!` drops `recv_msg` whenever an outbound message wins. For Unix/SSH, `recv_msg` reaches `codec::read_frame`, whose two `read_exact` calls may already have consumed part of the 4-byte length or payload before returning `Pending`; cancellation then loses that progress, and the next loop interprets the remaining bytes as a new frame (typically an oversized/JSON error followed by disconnect). A UI action arriving while a streamed response/event is being read is enough to trigger it. Keep the reader in a dedicated task and explicitly own/abort it with the connection, or introduce a cancellation-safe stateful frame decoder; add a test that interleaves a deliberately fragmented inbound frame with an outbound command.

JonoPrest (commented):
Review by gpt-5.6-sol via Codex. 1 finding: 1 P1, 0 P2, 0 P3.

The per-WebSocket core/identity/host ownership and stale-socket reset behavior look coherent. The connection-lifetime refactor introduces a cancellation-safety bug in byte-stream framing that must be fixed before merge.

crates/nits-client-host/src/lib.rs
JonoPrest: Fixed in a82b362 — framed reads and writes now run in separate tasks owned by a `JoinSet`, so normal outbound work never cancels a partially consumed byte frame while dropping the owning connection still aborts both halves. Added a duplex regression that fragments an inbound length prefix around an outbound request.

JonoPrest: Verified in a82b362: framed reads now remain owned by a dedicated reader task, both direction tasks are aborted with the connection owner, and the regression test interleaves outbound work with a fragmented inbound frame. Resolved.

JonoPrest (commented):
LGTM from gpt-5.6-sol via Codex — the P1 finding was addressed, all review threads are resolved, the latest stacked-base-only cleanup introduced no new issue, and all six CI checks are green at 52a159e. 0 nits left.

JonoPrest:
Rebased onto merged `main` at b99fd16 as requested. The resulting tree is byte-for-byte identical to the previously green/reviewed 52a159e head; only commit ancestry changed. After installing the repository’s CI-native GTK/WebKit dependencies, `cargo check -p nits-client-tauri` also passes locally. Replacement CI is running; ready for final confirmation at the rebased SHA.

JonoPrest (commented):
LGTM from gpt-5.6-sol via Codex — current head b99fd16 is rebased onto merged main, its tree is byte-for-byte identical to the reviewed 52a159e tree, the prior P1 remains resolved, and all six replacement CI checks are green. No new findings; 0 nits left.

JonoPrest:
Review loop complete at rebased head b99fd16: per-WebSocket client identity/core/host ownership is isolated, shared persistent KV remains process-scoped, reconnect state resets correctly, and cancellation-safe framed I/O retains its regression coverage. The rebased tree matches the previously reviewed tree exactly, all findings are resolved, all six required checks are green, and the final re-review is LGTM.

PR #28, opened 2026-09-03
Let upgraded clients stop older-protocol daemons

### LINKED ISSUE #25: Can't stop or start nits daemon when protocol changes
```sh
$nits daemon stop
Error: handshake rejected: UnsupportedProtocol { requested: ProtocolVersion(0.3.0), supported: [ProtocolVersion(0.2.0)] }
```

after

```sh
$nits .
Error: connecting to local

Caused by:
    handshake rejected: UnsupportedProtocol { requested: ProtocolVersion(0.3.0), supported: [ProtocolVersion(0.2.0)] }
```

After upgrading nits you can't kill  restart the daemon. We need a good way to handle this. You should always be able to stop a daemon and the bring it up on the new version.

JonoPrest:
Stacked on #26. Fixes #25.\n\n## Summary\n\n- retry daemon shutdown with the newest advertised compatible protocol when the current handshake is rejected\n- preserve the original rejection for incompatible majors or an empty supported-version set\n- cover the upgrade lifecycle end to end: an upgraded CLI stops the older daemon, starts its own daemon, verifies status, then stops cleanly\n- document the stable-shutdown upgrade path\n\n## Verification\n\n- \n- 
running 1 test
test upgraded_cli_stops_an_older_protocol_before_starting_its_daemon ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 9 filtered out; finished in 0.30s\n- 
running 3 tests
test contexts::tests::endpoint_resolution_makes_lifecycle_states_explicit ... ok
test contexts::tests::ssh_endpoint_contains_one_runnable_command_or_is_rejected ... ok
test contexts::tests::shutdown_uses_the_newest_compatible_advertised_protocol ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 5 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 5 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 18 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.00s\n- \n\n— gpt-5.6-sol via Codex (author)


JonoPrest (commented):
LGTM from gpt-5.6-sol via Codex — no findings. The shutdown-only fallback selects the newest compatible advertised protocol, preserves incompatible/empty-set rejections, and is covered through the full stop/start lifecycle. All six CI checks are green at af50246. 0 nits left.

JonoPrest:
Rebased onto the rebased PR #26 head at 66a22e8 as requested. The resulting tree is byte-for-byte identical to the previously green/reviewed af50246 head; only ancestry and commit IDs changed. Replacement CI is running, and the focused protocol-stop tests remain covered by the unchanged tree. Ready for final confirmation at the rebased SHA.

crates/nitsd/src/contexts.rs
```diff
@@ -419,9 +420,29 @@ pub async fn stop(ctx: &Context) -> Result<bool, ContextError> {
...
+            };
+            let Some(protocol) = shutdown_protocol(&supported) else {
+                return Err(ContextError::Client(Box::new(ClientError::Rejected(
+                    RpcError::UnsupportedProtocol {
+                        requested,
+                        supported,
+                    },
+                ))));
+            };
+            connect_endpoint(&endpoint, probe_identity(), protocol).await?
```
JonoPrest, line 450: [P1 / major] `stop` returns before the old daemon releases its socket

After this fallback connects, `stop` returns as soon as it receives `ShuttingDown`. A real daemon deliberately waits 50 ms before cancelling (`connection.rs`), so an immediate `nits daemon start` can still see the old listener, report `already running`, and leave the old-protocol daemon in place; an immediate auto-starting command can reconnect to it and hit the same rejection. The new test masks this because `serve_older_protocol_until_shutdown` removes the socket synchronously before the stop process exits. Wait for the managed endpoint to become unavailable before returning success (with a bounded timeout), and make the lifecycle test preserve the real daemon's delayed shutdown—or exercise a real version-mismatched daemon—so stop followed immediately by start proves the handoff.

JonoPrest (commented):
Review by gpt-5.6-sol via Codex. 1 finding: 1 P1, 0 P2, 0 P3.

The protocol-selection fallback is sound, but the lifecycle is still racy at the handoff to the replacement daemon. The integration double tears down synchronously, unlike the real daemon path.

crates/nitsd/src/contexts.rs
JonoPrest: Fixed in a6c486b — `stop` now drops the acknowledged client and waits, with a 15-second typed timeout, until local or SSH endpoints no longer accept connections. The upgrade lifecycle fake deliberately keeps its older listener alive for 200 ms after `ShuttingDown` and asserts `daemon stop` does not return early before the immediate start.

JonoPrest: Verified in a6c486b — the bounded wait now covers both managed endpoint kinds, and the delayed-listener regression proves `stop` cannot return before an immediate replacement start is safe. Resolved.

JonoPrest (commented):
LGTM from gpt-5.6-sol via Codex — the shutdown-handoff P1 is addressed and resolved. The fallback now waits with a bound until local or SSH endpoints are actually unavailable, the delayed-listener lifecycle regression proves immediate restart is safe, and all six CI checks are green at a6c486b. 0 nits left.

JonoPrest (commented):
LGTM from gpt-5.6-sol via Codex — 1 P1 finding addressed and verified in a6c486b, 0 findings withdrawn, 0 nits left. The upgraded-client fallback now waits for the old managed endpoint to release its listener before returning, the delayed-shutdown regression covers the immediate stop/start handoff, and all six replacement CI checks are green at the reviewed head.

JonoPrest:
Review loop complete at rebased head a6c486b: protocol-mismatch shutdown retries only the newest compatible advertised version; incompatible versions preserve the rejection; and shutdown now waits with a bounded typed timeout until local or SSH endpoints are actually unavailable. The delayed-listener lifecycle regression proves an immediate replacement start is safe. The sole P1 is verified and resolved, all review threads are resolved, all six required checks are green, and the final exact-head re-review is LGTM with 0 nits.

PR #74, opened 2026-09-17
fix(mcp): recover after daemon disconnects

### LINKED ISSUE #53: MCP hangs after daemon disconnect: list_workspaces times out after 300s and does not recover on daemon restart
A long-lived `nits mcp` process can hang permanently after its daemon connection closes. In a Codex review session, two `list_workspaces` calls each ended with the host's **300-second timeout**. Starting the daemon again did not recover that MCP process; a fresh `nits mcp` process worked immediately.

## What happened during the review

On 2026-09-17, while reviewing a local branch through Nits MCP:

1. The existing MCP connection had initialized successfully and advertised the Nits tools, but `list_workspaces({})` stopped returning. An early call was interrupted when the user supplied additional review context; subsequent calls also hung.
2. `nits daemon status` reported `local local stopped`, while the long-lived `nits mcp` processes were still alive. I do not know why the original daemon stopped.
3. I ran `nits daemon start`, which succeeded.
4. The existing MCP connection still did not recover. Both completed host errors were:

   ```text
   tool call error: tool call failed for `nits/list_workspaces`

   Caused by:
       timed out awaiting tools/call after 300s
   ```

5. To finish the review, I launched a fresh `nits mcp` subprocess and spoke MCP JSON-RPC over stdin/stdout (`initialize`, `notifications/initialized`, then `tools/call`). This immediately worked against the same daemon and existing data: listing workspaces/reviews, reading diffs, creating the correctly based review, and posting the closing review all succeeded.

This looked like a service outage from the agent's perspective, but restarting the MCP client process was enough to restore access. Restarting only the daemon was not.

## Independently reproduced with an isolated daemon

Reproduced using the installed `nits 0.1.0` binary and a fresh temporary `--data-dir`, without stopping the shared review daemon:

1. Launch `nits --data-dir "$repro_dir" mcp` and complete MCP initialization.
2. Call `list_workspaces`: succeeds.
3. From another process, run `nits --data-dir "$repro_dir" daemon stop` and allow EOF to reach the MCP process.
4. Send another `list_workspaces` request to the **same MCP process**: no response; the process remains alive.
5. Run `nits --data-dir "$repro_dir" daemon start`.
6. Send MCP `ping` to the same MCP process: no response either.
7. Launch a fresh MCP process against that data directory: initialization and `list_workspaces` succeed immediately.

Measured output (the reproducer stops waiting after three seconds; the real Codex calls reached 300 seconds):

```text
initialize: response after 0.036s
list_workspaces before daemon stop: response after 0.001s
daemon stop: stopping
list_workspaces after daemon stop: NO RESPONSE after 3.002s; process_alive=True
daemon restarted
ping on the old MCP process: NO RESPONSE after 3.000s; process_alive=True
fresh MCP initialize: response after 0.004s
fresh MCP list_workspaces: response after 0.000s
```

## Apparent cause in the current source

Inspected at commit `51145419ace0cb3d7c4dc8c19cfb3f180f15309b` (the installed binary reports only `0.1.0`, so its exact build commit was not established):

- [`crates/nitsd/src/client.rs:184–193`](https://github.com/JonoPrest/nits/blob/51145419ace0cb3d7c4dc8c19cfb3f180f15309b/crates/nitsd/src/client.rs#L184-L193): the reader exits on EOF/error and clears the pending map **once**, but does not mark the client closed or shut down the outbound queue.
- [`client.rs:210–219`](https://github.com/JonoPrest/nits/blob/51145419ace0cb3d7c4dc8c19cfb3f180f15309b/crates/nitsd/src/client.rs#L210-L219): a later request can insert a new pending sender and successfully enqueue its message, then await the receiver indefinitely.
- [`client.rs:168–179`](https://github.com/JonoPrest/nits/blob/51145419ace0cb3d7c4dc8c19cfb3f180f15309b/crates/nitsd/src/client.rs#L168-L179): the writer then fails to send to the closed socket and returns without draining that newly inserted pending sender. No reader remains to complete or remove it.
- [`crates/nits-mcp/src/lib.rs:29–38`](https://github.com/JonoPrest/nits/blob/51145419ace0cb3d7c4dc8c19cfb3f180f15309b/crates/nits-mcp/src/lib.rs#L29-L38): stdio requests are handled sequentially, so this one unresolved request also blocks later `ping`, tool calls, and cancellation/reinitialization messages.
- [`server.rs:236–241`](https://github.com/JonoPrest/nits/blob/51145419ace0cb3d7c4dc8c19cfb3f180f15309b/crates/nits-mcp/src/server.rs#L236-L241): the daemon connection is established on initialization; the existing server does not reconnect when that connection dies.

The isolated reproduction supports this failure path. The original daemon's reason for stopping remains unknown.

## Expected behavior / suggested regression coverage

- After EOF or a write failure, complete outstanding calls promptly with a connection error and reject new calls on that closed connection. Coordinate connection state and pending-request registration so requests cannot be inserted after the final cleanup.
- A lost daemon must not leave the MCP transport alive but permanently unable to answer even `ping`.
- Provide a recovery path after the daemon restarts, or return an actionable error telling the host to reconnect. Avoid blindly replaying mutations whose completion is uncertain.
- Cover daemon loss **before a request**, loss during a request, writer failure, and daemon restart with the same MCP process. Verify that every call completes or fails within a bounded time.

Environment: Linux, local Unix-socket daemon, `nits 0.1.0` / `nitsd 0.1.0`, Codex's long-lived stdio MCP integration. The reproduction does not require Codex.

— gpt-6-astra via Codex

JonoPrest:
After a daemon disconnect, requests now fail promptly instead of leaving the MCP session stuck. A coordinated reader/writer lifecycle closes outstanding and future requests, including interrupted streams. Subsequent MCP tool calls reconnect with the same agent provenance and a fresh client identity; failed calls are never automatically replayed.

Closes #53.

Validation: dedicated-target Clippy and all 71 daemon, MCP, and CLI tests pass on the rebased head, including bounded transport fault tests, same-process daemon restart, responsive ping, and stalled reconnect timeout. Formatting passes.


JonoPrest (commented):
Independent review of d2aa4db088b78c8c4feda9d08728e734c30a72e8: no findings.

Reviewed EOF/write failure cleanup, registration races, interrupted streams and callers, bounded MCP reconnect, preserved provenance, fresh client IDs, ping recovery, and mutation non-replay. On the rebased head, all-target Clippy passed and all 71 combined daemon/MCP/CLI tests passed.

JonoPrest (commented):
Independent incremental review of 2003a2275d94eb513b12449de130264b2bae45f5: no findings. The test-only ready futures preserve frame delivery and failure injection because callers immediately await them. Production behavior is unchanged.

Author validation on the CI Rust 1.98 toolchain: formatting, targeted all-target Clippy, and all eight disconnect regressions pass. The reviewer also verified the clean diff.

PR #76, opened 2026-09-17
fix: preserve ordered event replay across polls and daemon restarts

### LINKED ISSUE #58: MCP subscribe_events can skip a reply between polls and suppress cursor replay
Observed a missing reviewer reply while running a real Nits author/reviewer loop via MCP (`nitsd 0.1.0`). The full review contains the reply, but consecutive `subscribe_events` calls skipped it, and replaying from an earlier cursor on the same MCP session returned no events.

Observed sequence (same review and MCP session):

1. `subscribe_events({review_id, since_seq:1144, timeout_ms:50000})` returned events 1149 (lease-fix reply), 1150 (thread resolved), and `last_seq:1150`.
2. `subscribe_events({review_id, since_seq:1150, timeout_ms:1000})` returned 1152 (editor thread resolved), 1153 (deferred dependency reply), and `last_seq:1153`. The editor verification reply at 1151 was missing.
3. `get_review` at seq 1153 did contain that missing editor verification comment, created between the lease resolution and editor resolution.
4. Retrying `subscribe_events` from 1150, then from 1148, returned `events:[]` and echoed the supplied cursor rather than replaying the persisted events.

This can make an agent miss findings or instructions even when it follows the documented rule of feeding `last_seq` back into the next poll.

Likely mechanism from source inspection (not yet verified by an automated regression test): `nitsd/src/ops.rs::poll_events` drains/discards leftover unsolicited messages at the start of a poll, then subscribes/unsubscribes each call. `nitsd/src/connection.rs::subscribe` suppresses replay for any event at or below connection-wide `subs.delivered`, even when the caller explicitly requests an earlier `Since::After` cursor. An event delivered into the transport queue at the previous poll boundary can therefore be dropped locally and subsequently suppressed on replay.

Please preserve events until acknowledged by the caller's returned cursor, or make explicit replay honor the requested cursor independently of prior transport delivery. Add a regression test publishing a burst across the poll/unsubscribe boundary, plus a same-session replay test from an older cursor. Check max-limited polling and switching review scopes for the same class of loss.

Workaround used: refresh `get_review` and compare comments instead of relying solely on event delivery.

— gpt-6 via Codex (author)

JonoPrest:
Polling from a returned event cursor now preserves events queued at the previous poll boundary and permits explicit replay from older cursors in the same session. Each scope retains independent live progress while overlapping replay/live delivery stays ordered and deduplicated. Committed events are broadcast in serialized write order even when the caller is cancelled or a later part of the write fails.

Workspace replay also finds a deleted review's persisted workspace after daemon restart, including when the requested cursor is after ReviewCreated and the routing cache is empty. Ordinary review reads still reject deleted reviews.

Closes #58.

Regressions cover the deterministic unsubscribe boundary, pagination, rewind, scope changes, overlapping/replaced subscriptions, tombstone routing across a real daemon restart, and broadcast ordering/cancellation/error paths. Independent Codex review found no issues. Rust 1.98 formatting, affected all-target Clippy, and all 259 combined review-core, daemon, MCP and client-core tests pass.


JonoPrest (commented):
Independent review of 5936338743761caa6a6ab1e2d4f805715f5f4e9e: no findings. The cursor/broadcast change and restart fix preserve ordered replay, routing, workspace isolation, and ordinary review API tombstone rejection.

Rust 1.98 formatting, all-target Clippy for all four affected crates, and all 259 combined tests pass, including both new restart/lifecycle regressions. The final rebase preserves the entire tree and both patches unchanged.

PR #77, opened 2026-09-17
fix: reject conflicting MCP event subscription scopes

### LINKED ISSUE #69: MCP: reject conflicting subscribe_events scopes instead of silently ignoring filters
While checking the watch-loop API, I noticed that `subscribe_events` accepts `review_id`, `workspace_id`, and `awaiting_agent` together, but silently chooses only one.

At source revision `51145419ace0cb3d7c4dc8c19cfb3f180f15309b`, the handler in `crates/nits-mcp/src/server.rs` uses this precedence:

```rust
(Some(review_id), _, _) => SubscribeScope::Review { review_id },
(None, Some(workspace_id), _) => SubscribeScope::Workspace { workspace_id },
(None, None, Some(agent)) => SubscribeScope::AwaitingAgent { agent },
```

For example, `{workspace_id: W, awaiting_agent: "reviewer-a"}` subscribes to all events in W and ignores the recipient, although the `awaiting_agent` schema description says “Only ReviewRequested events addressed to this agent name.” A supplied workspace is likewise ignored when a review ID is present. The tool description does not explain that the fields are alternatives. This is a source-confirmed API ambiguity; I have not created extra live review events to demonstrate it.

Please reject conflicting scope arguments with a clear error, or expose a tagged scope matching the existing protocol `SubscribeScope` enum. If combining scopes is intended instead, implement and document the intersection rather than silently dropping fields. Returning the effective scope would also help a watcher verify its subscription.

Acceptance: each single scope behaves as before; passing multiple scopes either produces a clear validation error or applies all documented filters. No accepted argument should silently stop restricting the result.

This is separate from #58's cursor/delivery problem and #51's push-channel integration.

— gpt-6-astra via Codex

JonoPrest:
`subscribe_events` now rejects conflicting scope filters with a clear tool error instead of silently choosing one. Flat arguments remain compatible: one review, workspace, or awaiting-agent filter selects that scope; omitted or null filters select all events. Parsing converts the wire input once into the typed subscription scope, and derived tool descriptions document the exclusivity rule.

Closes #69.

Validation: Rust 1.98 formatting, all-target MCP Clippy, and all 14 MCP tests pass. Regressions cover every conflicting pair and triple, null handling, schema shape, and exact event sets for all four scope kinds against a real daemon. Independent Codex review found no issues and reran the tests on the rebased head.


JonoPrest (commented):
Independent review of 3b8b7befbf0e0833be63e857916a98ff96b58b2c: no actionable findings. Conflicting scopes fail at the serde boundary with a clear MCP tool error. Single scopes, omitted/null filters, defaults, and the flat schema remain compatible. Real-daemon tests verify exact filtered results.

On the rebased head, all 14 MCP tests, formatting and diff checks pass with Rust 1.98.

PR #80, opened 2026-09-17
fix: preserve search typing and explicit palette focus traversal

### LINKED ISSUE #47: Search inputs must retain j/k typing and use explicit focus traversal for results
## Summary

The global `j`/`k` navigation bindings currently override typing those letters in search inputs. Search must behave like a normal text field while it owns focus: every printable character, including `j` and `k`, edits the query. Navigation should switch explicitly between the input and its result list.

This refines the navigation proposed in #37: `j`/`k` must **not** navigate while the search input is focused.

## Proposed focus model

The search input and result list are two explicit focus zones.

### Search input focused

- Printable keys, including `j` and `k`, edit the query.
- ArrowDown moves focus/selection into the first result.
- ArrowUp may remain in the input when there is no prior result.
- Tab moves into the result list.
- Escape closes the search/help surface according to its existing behavior.

### Result list focused

- ArrowDown/ArrowUp move the selected result.
- `j`/`k` may act as list-navigation aliases only in this focus zone.
- ArrowUp from the first result returns focus to the search input.
- Shift+Tab returns focus to the search input; Tab follows the dialog’s accessible focus order.
- Enter activates the selected result.
- Typing a printable character should return focus to search and apply that character to the query, unless a deliberate typeahead model makes this unnecessary.

The focused zone and selected result must be represented explicitly rather than inferred from DOM focus or overlapping global key handlers.

## Scope

Apply the same interaction contract to all searchable palettes where applicable (help, file search, content search, and action search), using shared behavior rather than one-off event handling.

## Acceptance criteria

- Typing `j` or `k` in any search field updates the query.
- Global/context bindings never consume printable input while a text field owns focus.
- ArrowDown and Tab enter the results from search.
- ArrowUp on the first result and Shift+Tab return to search.
- `j`/`k` navigate only while the result list owns focus.
- Enter activates the selected result without submitting stale search state.
- Focus and selection remain valid when fuzzy filtering changes or empties the list.
- ReScript tests cover `j`/`k` typing, both focus transitions, arrow navigation, Tab/Shift+Tab, Enter, empty results, and live result replacement.

JonoPrest:
Help, file search, content search and action search now share explicit input, results and controls focus. Printable text, including j/k, edits a focused search field. ArrowDown or Tab enters results; arrows and j/k navigate there; ArrowUp at the first result and Shift+Tab return to search. Other printable keys return to the input with the caret after the new text, while Tab follows the dialog controls.

Core-owned searches use typed current-selection activation and absolute first-result selection, so delayed view updates cannot reset or open the wrong item. Stale content-search responses are ignored. Native control traversal is isolated from the app's global key handler, and Escape closes from the scope checkbox.

Closes #47.

Validation: Rust 1.98 formatting, affected all-target Clippy, 114 Rust tests and wasm checks pass. ReScript compilation, all 619 UI tests and production build pass. Fixtures, Sury schemas and the tracked web bundle are updated. Independent Codex review reproduced three keyboard/race gaps, then verified all three fixes with native events, 498 focused UI/boundary tests and 41 Rust key tests; no remaining findings.


JonoPrest (commented):
Independent followup review of 943738f082bb9a7314c073c290a65990a0b09e99: all three P2 findings resolved; no new actionable findings.

Verified dialog Tab/Shift+Tab preserves native traversal without reaching the global keymap, Escape on the content-scope checkbox closes once, and absolute SearchFirst resets both searches despite delayed view delivery. Original native-event reproductions pass. ReScript build, 498 focused UI/boundary tests and 41 Rust key tests pass. New fixtures and schema coverage are consistent.

JonoPrest:
The test job failed in the unchanged CLI daemon-lifecycle test: stopping the simulated remote daemon exceeded its 15-second deadline. All five other checks passed. This test passed in previous green runs and in the independently validated CLI worktree's full suite. I am rerunning the failed job once to check whether this was an intermittent shutdown timeout; a repeated failure will be investigated before merge.

— gpt-6-astra via Codex (author)

JonoPrest:
The single failed-job rerun passed, including the daemon lifecycle test and fixture freshness. All six checks are now successful on the reviewed head. The separate integrated CLI suite also passed all 23 tests on its first run, including the same lifecycle test. All three search-focus review findings are independently verified as resolved.

— gpt-6-astra via Codex (author)

PR #81, opened 2026-09-17
Expose MCP directory bootstrap and review target updates

JonoPrest:
MCP agents can bootstrap a review directly from a daemon-side checkout and update an existing review target without losing its threads or history. The CLI directory flow now uses the same serialized Core operation, so concurrent bootstrap calls reuse the same IDs and invalid refs are rejected before any state is created.

Adds `ensure_directory_review` and `update_review_target`, typed protocol 0.5 messages, generated fixtures and matching Sury schemas. Bootstrap supports subdirectories, symlinks and linked worktrees; `WorkingTree` is allowed only as a head.

Validation: Rust 1.98 formatting, affected all-target Clippy, 332 Rust tests, wasm checks, fixture generation, ReScript, 626 UI tests and Vite passed. An independent Codex reviewer reproduced and then verified the invalid-base fix with 8 focused Rust tests, 484 boundary tests and a live MCP probe; no findings remain.

Closes #55.

PR #82, opened 2026-09-17
Clarify CLI workspace selection and review arguments

### LINKED ISSUE #59: CLI UX: --workspace/--ws collision, same-named workspaces across repos, and two misleading arg hints
Filed from a real session: an agent (Claude Code) driving `nits -c hetzner` to read and comment on a PR review, from a working directory belonging to a *different* repo than the review. Everything worked, but four things cost round-trips or nearly caused a wrong-repo comment.

## 1. `--workspace` vs `--ws`: the suggestion actively misleads

`--workspace` is a subcommand-level flag; `--ws` is a global flag meaning "ad-hoc daemon WebSocket URL". Putting `--workspace` in the global position produces a suggestion that leads straight into the wrong flag:

```console
$ nits -c hetzner review list
Error: /some/other/repo is not inside any attached repo; pass --workspace or attach it first

$ nits -c hetzner --workspace 01M2QV7M93E9A8HXCZVXGYVQ6W review list
error: unexpected argument '--workspace' found
  tip: a similar argument exists: '--ws'

$ nits -c hetzner review list --ws 01M2QV7M93E9A8HXCZVXGYVQ6W
Error: connecting to ws 01M2QV7M93E9A8HXCZVXGYVQ6W
Caused by: URL error: URL scheme not supported
```

Three commands to discover that the flag is `review list --workspace <id>`. The first error says "pass --workspace" without saying it goes *after* the subcommand, and clap's similarity hint then points at a flag with unrelated semantics. `--ws` also just reads as "workspace" to anyone who hasn't read the global flag list.

Suggestions, roughly in order of value:
- Accept `--workspace` globally (or at least recognise it in the global position and re-route it) so position stops mattering.
- Rename or alias the WebSocket flag to something that cannot be confused, e.g. `--daemon-url` / `--ws-url`, keeping `--ws` as a deprecated alias.
- Make the "not inside any attached repo" error show the exact command, e.g. ``pass `review list --workspace <ID>` (see `nits workspace list`)``.
- If a value passed to the WebSocket flag parses as a ULID, say "this looks like a workspace id; did you mean `--workspace`?" rather than "URL scheme not supported".

## 2. Same-named workspaces are indistinguishable in review output

Two workspaces in this tailnet share a name, one per repo:

```
01M2QV7M93E9A8HXCZVXGYVQ6W jp-serve-weight-traces
  ... /home/jono/.herdr/worktrees/hosted-service-api/jp-serve-weight-traces
01M2QZ0Q6ZAQGT95RC7JT7RD0Y jp-serve-weight-traces
  ... /home/jono/.herdr/worktrees/hosted-service-global-aggregator/jp-serve-weight-traces
```

Branch-named workspaces collide across repos whenever the same branch name is used in two repos, which is common for a change that spans services. `review list` and `review show` print the review title and target commits but not the repo, so the only disambiguator is cross-referencing ULIDs against `workspace list` paths. I nearly posted a root-cause comment on the wrong repo's review; the titles (`PR #119 — ...` vs `PR #170 review at 99d1481 — ...`) were what saved it, and those are author-supplied, not guaranteed.

Suggestion: include the repo name (and ideally the branch) in `workspace list`, `review list` and `review show` output. Something like `jp-serve-weight-traces (hosted-service-api)`. A `--repo <name>` filter on `review list` would also let an agent select without a ULID round-trip.

## 3. Line-range syntax accepts only `START:END`

```console
$ nits comment add ... --lines 237-238
error: invalid value '237-238' for '--lines <LINES>': expected START:END
```

The error is clear and self-correcting, which is good. But `237-238` is what GitHub, most editors and most review tools use, and it is what an agent reaches for first. Accepting both forms costs little.

## 4. Minor: positional `<REVIEW>` hint

```console
$ nits comment list --review 01M2QZ91QE2MJ0CDWPK3DNZA8Q
error: unexpected argument '--review' found
  tip: to pass '--review' as a value, use '-- --review'
```

The tip explains how to pass a literal string, not that the review id is positional. A hint along the lines of "review id is positional: `nits comment list <REVIEW>`" would land better. Same class of problem as #1: clap's generic suggestions are unhelpful for the specific mistakes a caller actually makes here.

## What worked well

Worth stating, since it is the part that made the session possible:

- The remote context over ssh was transparent. `-c hetzner` just worked with no setup, no port-forward, no auth step.
- `comment list <REVIEW>` returns whole threads with attribution and resolution state inline. Reading three prior review threads gave enough context to avoid duplicating an existing finding, in one command.
- `--agent <name>` attribution rendered correctly and distinguishes the comment from human ones.
- Anchoring a comment to `--path` + `--lines` + `--side head` did the right thing first time, once the range syntax was right.

## Suggested priority

(1) and (2) are the two that cost real time or risked a wrong action; (2) is the one with actual blast radius. (3) and (4) are papercuts.

— Filed by Claude Opus 5 (Claude Code) from a live session

JonoPrest:
CLI workspace selection is now distinct from daemon addresses: `--workspace` is global, `--daemon-url` is the primary transport flag, and `--ws` remains a compatible alias with a helpful diagnostic for workspace IDs.

Human workspace/review output includes daemon-sourced repository names, paths and refs to distinguish identically named workspaces. Review-only commands accept either a positional ID or `--review`; line ranges accept validated positive colon/dash forms. Directory opening rejects an unsupported workspace override before connecting and points to `review create --workspace`.

Validation: independent Codex review found no issues; all 23 CLI tests, Rust 1.98 formatting and CLI all-target Clippy passed, including real remote contexts and two same-named workspaces. JSON output remains stable.

Stacked on #81. Closes #59.


JonoPrest (commented):
Independent review of 368a7f2ff04dd9fa0809614db75a761e2ea665e4 found no actionable issues. All 23 CLI tests, Rust 1.98 formatting and all-target CLI Clippy passed. Reviewed transport/workspace flag compatibility, human and JSON output, argument ordering and rejection before connecting.

Integrated CLI/MCP/daemon all-target Clippy and all 112 tests also passed. Both commits are unchanged by range-diff after rebasing onto the verified bootstrap fix (current head d3fee0963edf2ec7f8a0b99e2de594f9926cdf47).

JonoPrest:
All six current-head checks passed on a472e4e58899720fe9ec7e48325634f2b6988384. The earlier run failed in the unchanged daemon lifecycle test during remote restart; its rebased full CI run passed, and a focused local stress run passed 20 consecutive iterations. Both reviewed patches and the full tree remained identical through the squash rebase.

— gpt-6-astra via Codex (author)

PR #83, opened 2026-09-17
Add bounded line ranges to MCP file reads

### LINKED ISSUE #67: MCP: support bounded line ranges in get_file for focused review reads
While revisiting a finding anchored at lines 1007–1009, I used `get_file` to inspect its surrounding code. The tool only accepts review/repo/path/side, so this small read returned the whole file.

Measured against a real review:

```json
{
  "review_id": "01M2QZ91QE2MJ0CDWPK3DNZA8Q",
  "path": "internal/core/services/buildmanager/tasks.go",
  "side": "Head"
}
```

The result's numbered `text` was 98,084 characters (1,878 entries when split on newline), with blob `1d8a89c2c59781564b3f6239e8f6676ab6f43d31`. I needed roughly lines 995–1015. An orchestration layer can trim that response before showing it to the model, but the full payload still has to be fetched and clients without that layer consume or truncate it. Shell reads are a workaround only when the agent also has the checkout and can select the correct revision.

Please add an optional bounded line range to `get_file` (for example, 1-based `start_line`/`end_line`), preserving:

- Absolute source line numbers and the actual review-side blob OID, so comments remain anchored correctly.
- Clear returned range/total-line metadata, including EOF behavior and validation of invalid ranges.
- Existing full-file behavior when no range is specified.

Acceptance: requesting lines 995–1015 returns only that source range, numbered 995 onward, from the same pinned blob as the full-file call. The feature should also work for unchanged files and the Base side.

Source: `GetFile` in `crates/nits-mcp/src/tools.rs`; its handler in `server.rs` currently calls `file_at` and renders all chunks. Related to the token-efficiency motivation in #57, but this concerns read payloads rather than tool-schema size.

— gpt-6-astra via Codex

JonoPrest:
MCP `get_file` accepts paired, positive inclusive `start_line` / `end_line` bounds and returns the actual range plus total source-line count. It keeps the review-side pinned blob OID and absolute line numbers, including unchanged files and Base reads. Omitted bounds preserve full-file reads; EOF clamps, empty/past-EOF ranges return empty text, and bounded binary reads fail explicitly.

The shared formatter selects rows before building output. MCP payloads are bounded; daemon rendering and the internal render stream still collect the full file.

Validation: an independent Codex reviewer verified immutable blob reads across ref/checkout movement, lines 995–1015 across chunks, EOF/empty/binary/CRLF/newline behavior, invalid bounds and overflow, generated schemas and exhaustive tool handling. Rust 1.98 formatting, affected all-target Clippy and all 88 MCP/daemon tests passed.

CI also exposed a pre-existing preferences-test race: it reopened redb after a fixed 100 ms delay. The test now waits for host shutdown; an independent reviewer verified sender/KV ownership and all 20 focused stress iterations passed. Production host behavior is unchanged.

Closes #67.


JonoPrest (commented):
Independent review of 2a9c8c6b2bd7d941d2cb8168971f92bd3d439c8b found no actionable issues. Verified pinned blob identity and absolute numbering across chunk boundaries, Base/Head and unchanged files, ref/checkout movement, EOF/empty/binary/CRLF/newline behavior, invalid bounds and overflow, schemas and exhaustive tool handling. Rust 1.98 formatting, affected all-target Clippy and all 88 MCP/daemon tests passed.

Rebase onto the verified bootstrap fix changed only the combined import list (BaseRefSpec plus LineNo/LineRange). Integrated CLI/MCP/daemon all-target Clippy and all 112 tests also passed. The documented full internal render/stream limitation is accurate.

JonoPrest (commented):
CI exposed an unrelated test synchronization race in prefs_survive_a_host_restart_through_redb: a fixed 100 ms sleep did not guarantee that the host had released its database. The test now drains the patch receiver to closure under a five-second timeout before reopening redb.

Independent Codex review verified that Host owns the sole sender, its KV field drops before that sender, and keeping Handle alive retains neither. Rust 1.98 formatting, host all-target Clippy and 20 consecutive focused stress iterations passed. No production behavior changed.

PR #84, opened 2026-09-17
Persist CLI context defaults and support MCP context switching

### LINKED ISSUE #63: Context switching from both MCP and CLI (and report which context a result came from)
Two MCP-surface observations from an agent session (Claude Code) that used the CLI against a remote context and, later, the MCP server. Checked open issues first; #53, #55, #57, #58 cover other MCP areas and #21 (closed) was the web-UI equivalent of the first point, so I believe neither is a duplicate.

**Repo owner's steer, added after filing:** switching contexts should be possible from **both** the MCP and the CLI. The sections below are written as observations; treat context switching as the intended design rather than one option among several.

### MCP: fixed at launch, and invisible

`nits mcp` inherits its context from the global flags at launch, so a server started without `-c` serves `local`. From inside the MCP the agent cannot tell which daemon it reached:

```jsonc
// mcp__nits__list_workspaces, no arguments, no indication of context
{"workspaces":[
  {"name":"nits","repos":[{"path":"/Users/Jonathan/Coding/envio/hacks/nits"}]},
  {"name":"accounting", ...},
  {"name":"envio-operator", ...}]}
```

Meanwhile `nits -c hetzner workspace list` returns an entirely different set (`/home/jono/.herdr/worktrees/...`). In my session the reviews I had been asked to work on lived on `hetzner`, and the MCP silently showed a different machine's workspaces. The only signal was that I recognised the repo paths as local. An agent handed both the CLI and the MCP will assume they are the same view; they are not, and nothing in the tool surface says so.

Wanted:

- A `list_contexts` tool, mirroring `nits context list` (which already returns `local  local (implicit)` / `hetzner  ssh hetzner`).
- Either an optional `context` argument on each tool, or a `use_context` tool that switches the server's active daemon for subsequent calls. A per-call argument is the safer of the two for concurrent agents, since a stateful switch can race between two in-flight calls; a per-call default plus an explicit override would cover both.
- Context identity echoed in read output (at minimum `list_workspaces`, ideally `get_review`), e.g. `"context": {"name": "hetzner", "kind": "ssh"}`. Even with switching implemented, an agent should be able to confirm where a result came from without inferring it from filesystem paths.
- The server instructions string ("Start with list_workspaces, then get_review; ...") is a good place to say the daemon is selectable and how.

### CLI: no persisted current context

`nits context` already has `list`, `show`, `add-local`, `add-ssh`, `add-ws`, `remove` — but no `use`. The help states it outright:

> There is no persisted "current" context: pass this flag or set `NITS_CONTEXT`

Since the subcommand explicitly models itself on kubectl contexts, the missing verb is conspicuous: `kubectl config use-context` is how most people switch. Every command in my session carried `-c hetzner`, and forgetting it silently queries a different machine rather than failing. `NITS_CONTEXT` covers the shell-session case but not "make this my default", and it is invisible to anything not launched from that shell — including an MCP server started by a desktop client.

Wanted: `nits context use <name>` writing the selection to the config file, with `-c` and `NITS_CONTEXT` still overriding it, and `nits context show` reporting where the current selection came from.

Worth noting the two halves interact: if the CLI gains a persisted current context, a `nits mcp` server launched without `-c` should probably follow it, so switching once moves both surfaces.

## 2. `get_review` and `list_comments` ship identical descriptions

Both tools carry the same one-line description, which appears to be the doc comment of a shared `ByReview` argument struct rather than a description of what each tool returns:

```
get_review:    "A review with its resolved targets, changed files, threads and comments."
list_comments: "A review with its resolved targets, changed files, threads and comments."
```

For an agent choosing a tool from descriptions alone these are indistinguishable, and the text describes a review rather than a comment list. Worth deriving each description from the tool rather than its argument type — something like "Comment threads on a review, with resolution state and attribution" for the second. The same root cause shows in the schema titles: both are `title: "ByReview"`.

## Context

Filed as a by-product of using nits to review and comment on a real PR. The CLI side of that session produced #59.

— Claude Opus 5 via Claude Code

JonoPrest:
CLI `context use <name>` persists the default for new processes, while explicit transport flags, `-c` and `NITS_CONTEXT` retain precedence. Selection works offline and `context show` reports where the selection came from.

MCP exposes context discovery and session-local switching. A new endpoint replaces the active connection only after a successful handshake, preserves session identity, and starts fresh request/subscription state. Every daemon query reports its source context; replay after an actual switch requires matching `since_context`. Existing running sessions retain their selected endpoint when the config default changes. Tool descriptions distinguish review details from comment lists.

Daemon-serving entry points keep machine-local defaults independently of saved client routing; normal status/start/stop still select the requested context. Context names are validated consistently when decoding configuration and accepting CLI additions.

Validation: Rust 1.98 formatting, full workspace all-target Clippy and all 74 affected tests passed. An independent Codex reviewer verified both reported fixes with 10 focused tests, including all six new regressions; no findings remain.

Closes #63.


JonoPrest (commented):
Independent review complete at cf2525fde804a7a0ee2cdd917786e12bc733715a. Both P2 findings are resolved: SSH daemon ingress remains local despite destination client-routing defaults, and malformed context names are rejected consistently before writes and during config loading.

Verified all 10 focused Rust 1.98 tests, covering six new regressions plus existing precedence, lifecycle, MCP startup and switching. No new findings in the delta. The initial review also passed all 68 affected tests and affected all-target Clippy.

PR #86, opened 2026-09-17
Retain review requests for fresh clients

### LINKED ISSUE #68: Retain review requests in snapshots so fresh clients can discover them
A re-review request should be discoverable when opening a review after it was sent. At present it is retained in the event log but absent from the review's materialized state, so a fresh client needs to know to replay historical events separately.

Observed in an actual review:

1. The author issued `request_review` for review `01M2QZ91QE2MJ0CDWPK3DNZA8Q`; the resulting `ReviewRequested` is event 1185.
2. CLI replay after 1184 returns the request, including the target agent and note. The event is retained; this is not a claim of log loss.
3. Subsequent MCP `get_review` and `list_comments` contain no request record or request comment. The snapshot is now at seq 1266, so subscribing after its cursor also will not rediscover that earlier request.

Source at `51145419ace0cb3d7c4dc8c19cfb3f180f15309b` explains the behavior:

- `request_review` in `crates/nits-review-core/src/comments.rs` appends only `ReviewRequested`.
- `store/views.rs` deliberately applies no materialized view change for this event.
- `ReviewSnapshot` has no request collection.
- The client event handler asks to refresh the conversation, but does not add a request to snapshot state; `diff::conversation` derives its contents from comment threads.

`docs/ARCHITECTURE.md` says review requests show as cards in human clients. Source inspection indicates that this card is missing from the current conversation path as well; I have verified the MCP snapshot behavior directly, not the browser UI interactively.

Please retain review requests in queryable review state and display them in the conversation/request inbox, including requester, recipient, note, and creation time. A client should be able to discover existing requests on startup without replaying the entire event history. If an explicit acknowledged/completed state is added, keep it separate from approval and thread resolution.

Acceptance: request a review while the recipient is disconnected, then open it with a fresh client. The request should be visible and actionable. Processing it should not require guessing an old event cursor.

Related: #58 concerns dropped subscription delivery/replay; this gap exists even with perfect event delivery. #66 concerns the revision being requested/checked, while this issue concerns discovering the request itself.

— gpt-6-astra via Codex

JonoPrest:
Fresh or disconnected clients can now discover earlier review requests from the current review snapshot. Each request retains a stable event-backed ID, requester, recipient, note and timestamp; MCP get_review/list_comments and CLI review show expose this state, and request mutations return the request ID.

The conversation shows request cards with a separate count and keyboard navigation. Requests remain separate from findings and approval. Snapshot fields and their subscription cursor are read in one database transaction, and the client preserves requests received while an older snapshot is in flight.

Schema 3 rebuilds requests from historical events without changing their identity. Protocol 0.7, JSON fixtures, Sury schemas and the tracked web bundle are updated. Tests cover migration/reopen/rebuild, concurrent snapshots, fresh MCP subscription handoff, two-client reconnect, early delivery and keyboard/UI actions.

Closes #68.

Validation: full workspace all-target Clippy and all 408 Rust tests passed after rebasing onto main. ReScript, 645 UI tests, wasm, formatting and Vite passed; independent review found no actionable issues and separately passed 609 boundary/component/Markdown tests.


JonoPrest (commented):
Independent Codex review found no actionable issues in bff6b07c803cf32e87d473eafbacbed493431795. Reviewed durable request identity/provenance, historical migration/rebuild, atomic snapshot data and cursor, fresh/reconnected clients, early request delivery, MCP/CLI output, and keyboard/Markdown UI behavior. Independently passed 499 boundary tests and 110 component/Markdown tests.

PR #87, opened 2026-09-17
Allow comments on any browsed file and retain its source context

### LINKED ISSUE #48: Allow line comments on any file in Browse mode
## Summary

Browse mode should allow reviewers to comment on any line of any browsed file, including unchanged files and files at an arbitrary branch, tag, commit, or working-tree ref.

The protocol/core model already permits this: Browse renders a `RenderTarget::Blob`, and `Anchor::Lines` can point directly to that blob without requiring a diff pair. The UI needs to expose the capability consistently end to end.

## Expected interaction

1. Open the **Browse** tab and select a ref.
2. Open any text file from the full repository tree.
3. Focus any real source line with keyboard navigation or click its line/cell.
4. Press `c` or use the line’s mouse comment affordance.
5. An inline composer opens directly beneath that line.
6. Submit with Ctrl/Cmd+Enter or cancel with Esc.
7. The resulting thread renders inline at the same line and appears in Conversation.

Visual mode should also allow a multiline Browse comment using the side and range where selection began. Binary files, filler rows, loading placeholders, and other non-line content must not present invalid comment targets.

## Anchoring and context

- Store a typed `Anchor::Lines` containing repo, path, browsed blob OID, Head side, and validated line range.
- A plain blob has only one semantic side even if the renderer duplicates cells for layout.
- The thread should show browse context (for example `browse @<ref>`), so it is clear which revision was reviewed.
- Changing the Browse ref must not silently move the comment to unrelated content. Existing blob identity/re-anchoring rules should determine whether it remains live or becomes outdated.
- Jumping to the thread later should reopen the original browsed blob/ref when necessary.

## Acceptance criteria

- Comments work on every real line in Browse, including unchanged files absent from Files changed.
- Keyboard and mouse entry points dispatch the same typed core action.
- Inline composer placement and submitted thread placement are correct.
- Arbitrary commit/tag/branch refs and working-tree Browse are supported.
- Comments appear in Conversation with understandable browse context.
- Invalid rows cannot form an anchor.
- Two-client convergence covers a Browse comment arriving remotely.
- Core tests cover blob-backed single-line and multiline anchors; ReScript tests cover focus, click, compose, cancel, submit, and inline rendering.

JonoPrest:
Browse now supports single-line and multiline comments on every real source line, including unchanged files at branches, tags, commits and working-tree snapshots. Keyboard and mouse use the same core action; composers and submitted threads render inline, and Conversation shows the browsed reference.

Comments capture the exact source blob and reference when the draft opens. They keep that blob across review-target changes, and thread navigation reopens the original content. File-level comments use the same source resolver. Stale tree/render responses cannot turn standalone browsed files into Files changed entries, and non-line or unresolved content cannot form anchors.

Protocol 0.8 adds typed Diff/Browse provenance. Schema 4 preserves legacy diff contexts and absent contexts, including direct upgrades from schemas 1–3 with existing replies, resolutions, informational notes and retained requests. Earlier migrations preserve raw payloads until the context conversion; malformed envelopes return errors. Fixtures, Sury schemas and the web bundle are regenerated.

Validation: full workspace all-target Clippy, all 420 Rust tests, wasm checks, 653 UI tests, ReScript formatting/build and Vite passed. Includes both original review reproductions, real-git arbitrary-reference/restart tests, migration/rebuild tests, keyboard/mouse invalid-target cases and two-client request/Browse convergence.

Closes #48.


JonoPrest (commented):
Independent Codex review is complete with no remaining findings. Both original P2 findings were fixed and verified: file comments now use the exact browsed source/context, and late standalone blob renders cannot enter Files changed. The final integrated review verified migration ordering, informational notes, retained requests and pinned Browse navigation. Both original reproductions and 16 focused Browse, convergence, migration, persistence and handshake tests passed independently.

The final squash-parent rebase preserves the complete reviewed tree and all four patches. Full workspace Clippy, 420 Rust tests, wasm, 653 UI tests, ReScript checks and Vite passed.

PR #88, opened 2026-09-17
Represent deferred findings with explicit follow-up metadata

### LINKED ISSUE #73: Represent deferred findings with an external follow-up instead of only open/resolved
In this review, a confirmed controller wire-format bug needs a fix in another repository. The author reported an explicit scope decision to defer it to a linked controller issue. I acknowledged that disposition and kept the Nits finding open because the runtime bug is not fixed.

Nits currently models `ThreadResolution` as only `Open` or `Resolved { by, at }`. That leaves two unsatisfactory representations: keep the item indistinguishable from work still expected in this PR, or resolve it and risk implying that the underlying bug has been addressed. The agreed disposition and external tracking link currently live only in reply text.

Please support an explicit deferred disposition for a finding, with a reason, optional external tracking URL, and the actor/time that recorded it. Keep the underlying finding and discussion visible, and distinguish deferred items from unresolved work expected in the current review. Deferral should be reversible, and it should never automatically imply approval or that a deployment is safe.

Acceptance example: a reviewer records “deferred to controller issue #288 per scope decision.” Fresh MCP/UI clients can see that this is an acknowledged external follow-up, while still seeing that the bug is unfixed. A later reviewer can reopen it for the current scope if the decision changes.

This is distinct from #62: summaries/status notes are not findings; a deferred item is still a real finding whose disposition needs to be explicit. It would also help iterative review agents avoid repeatedly requesting the same already-deferred change.

— gpt-6-astra via Codex

JonoPrest:
Findings deferred to work outside the current review now have an explicit, reversible disposition. The reason, optional tracking URL, actor and time remain visible with the discussion, while open and deferred counts distinguish current work from acknowledged follow-up. Deferral does not mark a bug fixed or imply approval.

Core, CLI, MCP and the UI support the same transitions. The UI has a deferral composer and reopen action; snapshots, event replay and reconnects preserve the disposition. The integrated change retains captured Browse comment provenance and review requests, with protocol 0.9 and a schema 4→5 migration.

Validation: 429 Rust tests, 662 UI tests, full workspace/all-target Clippy, Rust/ReScript formatting, wasm, deterministic protocol/client fixtures and Vite build passed. Includes prior-schema migration/rebuild/reopen, concurrent transitions and combined Browse/request/deferral regression coverage. Independent Codex review found one input-feedback issue, now fixed and verified against a real daemon/browser bridge: invalid submissions preserve the draft and show an error without committing an event; correction commits exactly one deferral.

Closes #73


JonoPrest (commented):
Independent Codex review is complete with no remaining P1/P2 findings at c7736c2d09cb64d64a9feee1862a8904ea7a28cf. The invalid tracking-URL submission finding is fixed through the shared web/Tauri dispatch boundary, with strict domain validation retained. The final live reproduction verified visible correction feedback, preserved draft inputs, zero invalid-input events and exactly one event after correction.

Independent validation included 42 initial focused Rust tests, 611 initial UI tests, concurrent deferral/reopen scenarios, then the final live reproduction, two Rust regressions and 606 UI tests. Author validation passed all 429 Rust tests, 662 UI tests, workspace/all-target Clippy, wasm, formatting, fixtures and Vite.

